PR #2, opened 2019-02-20
I2610 - dettl can use vault secrets for connecting to DB

r-ash:
* Copied code (and test) from orderly which manages reading vault secrets
* Added tests to check that we can use these to access a remote DB using creds from vault
* Added notes to README pointing to setup information about using vault


R/vault.R
```diff
@@ -0,0 +1,33 @@
...
+#'
+#' @param x List of properties some of which may be vault secrets of the form
+#' VAULT:<path to secret>:<key to get>
+#' @param config Cnfig containing details of the vault server to be accessed.
+#'
+#' @return List of properties with any secrets resolved secret.
+#'
+#' @keywords internal
+#'
+resolve_secrets <- function(x, config) {
```
richfitz: If we're using this in both `orderly` and in `dettl`, do you think it should go into `vaultr` perhaps?

R/dettl_config.R
```diff
@@ -46,7 +46,7 @@ dettl_locate_config <- function(path) {
 #'
 db_config_read_yaml <- function(filename, path) {
   info <- yaml_read(filename)
-  check_fields(info, filename, "source", "destination")
+  check_fields(info, filename, "source", c("destination", "vault_server"))
```
richfitz, line 49: It is probably worth validating the type of the `vault_server` field when present

tests/testthat/test-db.R
```diff
@@ -1,12 +1,30 @@
...
+test_that("dettl can connect to remote DB using yaml config and the vault", {
+  srv <- vaultr::vault_test_server()
+  cl <- srv$client()
+  cl$write("/secret/users/readonly", list(password = "readonly"))
+
+  path <- tempfile()
+  dir.create(path)
+  filename <- file.path(path, "db_config.yml")
+  cfg <- readLines("uat_config/db_config.yml")
+  cfg_server <- gsub(pattern = "<vault_server>", replace = srv$addr, x = cfg)
```
richfitz: Tiny thing, but I would be inclined to used `fixed = TRUE` here also

richfitz (changes requested):
Main thing is wondering if we should move the secret resolution bit into `vaultr` itself - what do you think?

R/vault.R
r-ash: I think that would be a good idea. It it identical between `orderly` and `dettl`.

r-ash:
Okay I think is ready for another look. I've addressed the points raised including moving the work to resolve secrets into `vaultr`. I've also expanded the test coverage (which is now working again after the tests have been rewritten) .

richfitz (approved):

PR #3, opened 2019-02-22
I2678 - Update db config to allow running import on separate databases

r-ash:
* Can now specify the DB you want to run the import on when you create a new `dataImport` object via `new_import` method. Returns a useful message if the db specified is not configured.
* Updated vignette
* Also fixed an issue where pkgdown website couldn't be built due to an bug with tearing down a db used for testing within context of the vignette.


R/dettl_runner.R
```diff
@@ -1,11 +1,13 @@
...
 #'
 #' @param path Path to directory containing functions for import process.
+#' @param db_name The name of the db to connect to. Connection info must be
+#' configured via the `db_config.yml`.
 #'
 #' @return A dataImport object.
 #' @export
 #'
-new_import <- function(path) {
+new_import <- function(path, db_name) {
```
richfitz: in `orderly` we default to the first listed name in the configuration (so take a `NULL` argument here). What do you think?

R/test-helper.R
```diff
@@ -2,8 +2,11 @@
 prepare_example_db <- function(db_name, dir = getwd()) {
   path <- file.path(dir, db_name)
   con <- DBI::dbConnect(RSQLite::SQLite(), path)
+  if (DBI::dbExistsTable(con, "people")) {
```
richfitz: is this because of modifying test data in place? I think you'd be better off in the long creating things freshly, or you'll get weird test failures that depend on the order tests are run in

R/dettl_runner.R
richfitz: Also, and in a different ticket probably, `new_import` is an odd name - it sounds like something you might run to create the skeleton of a new import (we have this in `orderly` as `orderly::orderly_new`). Not sure what a better name would be though

R/dettl_config.R
```diff
@@ -46,28 +48,24 @@ dettl_locate_config <- function(path) {
...
 
-  driver_config <- function(name) {
-    if (name == "destination" && is.null(info[[name]])) {
-      ## Fall back to default destination DB
-      info[[name]] <- list(driver = "RSQLite::SQLite",
-                           args = list(dbname = "dettl.sqlite"))
+  check_length(info$db, "gt", 0)
+  for (db_cfg in names(info$db)) {
+    if (is.null(info$db[[db_cfg]]$driver)) {
+      stop(sprintf("No driver specified for DB config %s.", db_cfg))
```
richfitz, line 56: this line is untested

richfitz (changes requested):
This looks like a good way forward - there are a few notes below (I've not run this yet locally though!)

You don't test the

R/dettl_runner.R
r-ash: Using a default is a good idea, I'll do that. 

Yeah noted the point about the `new_import` I was probably thinking about what the method was doing instead of thinking from user perspective. They're initialising an or setting up an import for a directory and a database so `setup_import` or `initialise_import` is probably more appropriate. I can add a separate ticket for that as you suggested.

R/test-helper.R
r-ash: So this wasn't a problem in the tests, everywhere in the tests I'm cleaning the database up using `on.exit(unlink(path_to_db))`. 

It was causing problem when knitting the vignette as the `people` table I'm creating in this setup would already exist. I'm not fully sure why it was failing to clear up the file when running in vignette. I see the point though, I can make this method just create the whole test db from new.

richfitz (approved):

PR #5, opened 2019-03-12
i2752 - Make connection available to tests

r-ash:
This makes connection available to the extract, transform and load tests

Generally expect we will want to inspect the database to make sure that the load stage has passed successfully. 

More specifically in porting example data we have instances where we want to access the db during the transform stage to read fks for use in data validation.

### DIFF
--- a/R/extract_runner.R
+++ b/R/extract_runner.R
@@ -20,7 +20,7 @@ run_extract <- function(con, extract, path, extract_test){
     stop("DB connection is not valid cannot extract data")
   }
   if (!is.null(extract_test)) {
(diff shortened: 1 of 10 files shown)


richfitz (approved):
Looks good @r-ash - based on the number of places that are changing at once I wonder if a small data object that pulls the various components that always go together might simplify things?

r-ash:
> Looks good @r-ash - based on the number of places that are changing at once I wonder if a small data object that pulls the various components that always go together might simplify things?

This would be some object which makes the extracted data, transformed data and connection available? I'd then pass that to each of the test runners instead?

PR #8, opened 2019-04-04
Improve messages around testing

r-ash:
This PR will
* Print a message when any of the extract, transform or load tests are about to be run
* Print a message when any of them have been successfully run
* Tested that that is the case
* Also take the opportunity to switch to using `summary` reporter in the vignette to avoid escape characters used to control colouring by the default `progress` reporter. 
   * On this point I tried pretty much every reporter and it looks like `summary` is the best option available to me here (no colouring but still printing some useful information but not too verbose). Though it's still not perfect in that it prints a large horizontal line of `=` for formatting which doesn't look great in the vignette. I'm not sure if there is a good way to work around that?

### DIFF
--- a/tests/testthat/test-transform-runner.R
+++ b/tests/testthat/test-transform-runner.R
@@ -61,3 +61,30 @@ testthat::test_that("verification passes if data adheres to schema", {
   colnames(transformed_data$people) <- c("name", "age", "height")
   expect_silent(verify_data(con, transformed_data))
 })
+
+context("load_runner")
+
+testthat::test_that("messages are printed to console when tests are run", {
+  transform_func <- function(data, con) {
+    list(people = stats::setNames(
+      c("Test", 2, 3),
+      c("name", "age", "height")
+    ))
+  }
+  db_name <- "test.sqlite"
+  prepare_example_db(db_name)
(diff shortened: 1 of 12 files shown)


r-ash:
On second thought regarding the last point we discussed detecting whether the test runs are printing to a terminal or not and using this to manage the reporter that is used. I can probably handle that in a separate PR.

tests/testthat/test-transform-runner.R
```diff
@@ -61,3 +61,30 @@ testthat::test_that("verification passes if data adheres to schema", {
   colnames(transformed_data$people) <- c("name", "age", "height")
   expect_silent(verify_data(con, transformed_data))
 })
+
+context("load_runner")
```
richfitz: one context per file - please move to another one (some versions of test that have had odd behaviour with this)

richfitz (changes requested):
LGTM - just one bit of test reorganisation needed

tests/testthat/test-transform-runner.R
r-ash: I've removed that line, I think probably a copy + paste error which should never have been there.

richfitz (approved):

PR #9, opened 2019-04-04
Add ability to use glob patterns for source configs

r-ash:
This PR will
* Support sourcing files using pattern matching


R/dettl_config.R
```diff
@@ -189,10 +189,40 @@ read_fields <- function(fields, config, env) {
...
+#'
+#' @keywords internal
+expand_wildcards <- function(sources, path) {
+  all_files <- c()
+  for (source in sources) {
+    files <- Sys.glob(file.path(path, source))
+    if (length(files) > 0) {
+      all_files <- c(all_files, files)
+    } else {
+      message(sprintf("No files found matching file pattern %s", source))
```
richfitz: I feel that this should be an error - if not then you could just use:
```r
expand_wildcards <- function(sources, path) {
  Sys.glob(file.path(sources, path))
}
```

R/dettl_config.R
```diff
@@ -189,10 +189,40 @@ read_fields <- function(fields, config, env) {
...
+#'
+#' @param sources List of configured file paths some of which may contain
+#' wildcards. Relative to the root of the import.
+#' @param path Normalised path to the import directory.
+#'
+#' @return List of files to be used as sources
+#'
+#' @keywords internal
+expand_wildcards <- function(sources, path) {
+  all_files <- c()
```
richfitz: This pattern (growing a vector) is not very R-ish: more typical would be

```r
expand1 <- function(x) {
  files <- Sys.glob(x)
  if (length(files) == 0) {
    message(sprintf("No files found matching file pattern %s", x))
  }
  files
}

expand_wildcards <- function(sources, path) {
  unlist(lapply(sources, expand1, path), FALSE, FALSE)
}
```

R/dettl_config.R
```diff
@@ -189,10 +189,40 @@ read_fields <- function(fields, config, env) {
...
+  for (source in sources) {
+    files <- Sys.glob(file.path(path, source))
```
richfitz: If you want to separate out the path part from the pattern matching, the other option is to use `withr::with_dir` to change working directory before processing the patterns - this usually results in cleaner code

richfitz (changes requested):

r-ash:
Thanks for the feedback above. I've addressed each of those points.

richfitz (approved):

PR #10, opened 2019-04-09
Add config for foreign key constraints

r-ash:
This PR will
* Allow users to run an import using default load step by initialising the import using `dettl("path", default_load = TRUE, db_name = "uat")`
* Improve method naming, to initialise an import use `dettl` instead of `new_import`
* Allow users to configure foreign key constraints for use in `load` stage in the `dettl.yml` config
* Update vignette to include information about how to use this


tests/testthat/test-run-default-load.R
```diff
@@ -0,0 +1,53 @@
...
+    DBI::dbClearResult(query)
+    DBI::dbGetQuery(con, "SELECT last_insert_rowid()")[1, 1]
+  }
+  run_import(import, c("extract", "transform"))
+  data <- import$get_transformed_data()$people
+  mock_insert <- mockery::mock(insert_func(data[1, c("name", "age", "height")]),
+                               insert_func(data[2, c("name", "age", "height")]))
+
+  ## Run load with mock INSERT function
+  with_mock("dettl:::insert_data" = mock_insert, {
```
r-ash: Here I am mocking the code which does the actual insert as that uses an `INSERT INTO RETURNING` pattern which is not available in SQLite. This looked like the best way to work around it though I'm not super happy with it as it results in lots of code in the test and means I'm not testing the entire `get_default_load` function. I wonder if that is just something we have to accept as the test driver is using SQLite instead of Postgres so there will be some differences. 

I'm hoping the explanation lines 15-22 when I start writing the mocking is enough to make this clear too?

R/dettl_runner.R
```diff
@@ -1,24 +1,31 @@
 #' Create an import object using functions defined at specified path
 #'
 #' @param path Path to directory containing functions for import process.
+#' @param default_load If true then load step will use default implementation.
```
richfitz: I am not sure why this is an argument - this should be a property of the configuration, not the evaluation in my mind?

R/dettl_runner.R
```diff
@@ -1,24 +1,31 @@
...
 #' @param db_name The name of the db to connect to. Connection info must be
 #' configured via the `db_config.yml`. If name is left blank this will default
 #' to using the first db configured.
 #'
-#' @return A dataImport object.
+#' @return A DataImport object.
```
richfitz, line 8: We should document up somewhere how we feel about class names and the like

tests/testthat/test-dettl-runner.R
```diff
@@ -13,11 +13,11 @@ test_that("dettl works as expected", {
   on.exit(options(testthat.default_reporter = default_reporter), add = TRUE)
 
   ## when creating import object
-  import <- new_import("example/", "test")
+  import <- dettl("example/", db_name = "test")
 
   ## object has been created
   expect_false(is.null(import))
-  expect_equal(class(import), c("dataImport", "R6"))
+  expect_equal(class(import), c("DataImport", "R6"))
```
richfitz: you can also use `expect_is` here to to check class attributes

tests/testthat/test-foreign-key-constraints.R
```diff
@@ -0,0 +1,72 @@
+context("foreign-key-constraints")
```
richfitz, line 1: Putting the comment here because I don't know where it would otherwise naturally go - as Alex found out the other day, one has to opt in to SQLite's FK enforcement.  See

https://github.com/vimc/orderly/blob/master/R/db2.R#L568-L572

which we use when creating the connection

tests/testthat/test-run-default-load.R
richfitz: I'm confused why you have both `mockery`'s mock and testthat's `with_mock` here

tests/testthat/test-run-default-load.R
```diff
@@ -0,0 +1,53 @@
...
+  ## Turn off reporting when running import so import tests do not print
+  ## to avoid cluttering up test output.
+  default_reporter <- testthat::default_reporter()
+  options(testthat.default_reporter = "silent")
+  on.exit(options(testthat.default_reporter = default_reporter), add = TRUE)
+
+  import <- dettl("example_default_load/", default_load = TRUE,
+                  db_name = "test")
+
+  ## Setup mock INSERT query for testing. This is needed because in real
```
richfitz: I think that a possibly better solution would be to abstract this bit of logic into a function `insert_into_returning` (or whatever!) that switches based on the SQL dialect.  So if you generalise the `insert_values_into` function so that the form you have is for postgres, then add another one for sqlite (you could do this with either a switch or S3 methods perhaps)

R/default_load.R
```diff
@@ -0,0 +1,117 @@
+## impl and test
```
richfitz: I think a high level overview of the approach here in a comment would be useful

richfitz (changes requested):
* rather than dealing with the sqlite/postgres differences by mocking, I think it would be preferable to fix this by dialect switching in the package itself.  At some point we will want to use `dettl` on a non-postgres db I am sure (a SQLite one or a MySQL one).  I don't think that will massively inflate the PR further, and might make it easier to test.  In order to test on travis you can request a postgres db that you could use for testing that function and we can get you going with something like docker for a throwaway db for testing locally.
* I think that the auto load configuration is in the wrong place and that it should be a part of the import not part of the invocation of the import
* At present you require the user to fill in the fk constraints, but I wonder if it would be simpler to have the DB do that for us (that would save at least the configuration bits).  For example, see https://dba.stackexchange.com/questions/36979/retrieving-all-pk-and-fk - quickly checking with montagu shows

```sql
SELECT c.conname                                     AS constraint_name,
       c.contype                                     AS constraint_type,
       sch.nspname                                   AS "schema",
       tbl.relname                                   AS "table",
       ARRAY_AGG(col.attname ORDER BY u.attposition) AS columns,
       pg_get_constraintdef(c.oid)                   AS definition
FROM pg_constraint c
       JOIN LATERAL UNNEST(c.conkey) WITH ORDINALITY AS u(attnum, attposition) ON TRUE
       JOIN pg_class tbl ON tbl.oid = c.conrelid
       JOIN pg_namespace sch ON sch.oid = tbl.relnamespace
       JOIN pg_attribute col ON (col.attrelid = tbl.oid AND col.attnum = u.attnum)
GROUP BY constraint_name, constraint_type, "schema", "table", definition
ORDER BY "schema", "table";
```

returning really everything we need.  There will be an equivalent to this for sqlite - e.g. https://stackoverflow.com/questions/5499003/sqlite-list-all-foreign-keys-in-a-database

tests/testthat/test-run-default-load.R
r-ash: This was following the examples from `mockery` documentation for `mock` and the "simple integration" section mockery [vignette](https://cran.r-project.org/web/packages/mockery/vignettes/mocks-and-testthat.html). Is there a mockery equivalent or `with_mock`?

I could avoid using the `mockery::mock` as I am not doing any interaction testing but I think that would make the mock function more complicated.

R/dettl_runner.R
r-ash: So I updated this after seeing Hadley Wickham's recommendation in Advanced R
> By convention, R6 classes have UpperCamelCase names

Where you thinking documenting this at a repo level or something more general?

richfitz: more generally - we don't have a style guide we can point to yet

tests/testthat/test-foreign-key-constraints.R
r-ash: I've pushed https://github.com/vimc/dettl/pull/10/commits/6839705aae34476d2e9454342aa2f153063a08c3 to address this

r-ash:
Okay I think this addresses all the points raised, in particular
* Now flag for using default load is in the `dettl.yml` config
* Add new generic function for inserting values in DB returning id of added data and impl for SQLite and Postgres
* Better documentation

I've created a separate ticket for moving away from having to configure the foreign key constraints and instead reading them from the DB.

R/db.R
```diff
@@ -7,11 +7,28 @@
...
 }
 
+#' Create a connection a DBMS.
+#'
+#' Uses \code{\link[DBI]{dbConnect}} to connect to a DBMS. If this uses
+#' \code{\link[RSQLite]{SQLite}} then it will ensure foreign key constraints are
+#' enabled.
+#'
+#' @keywords internal
+dbi_db_connect <- function(drv, ...) {
```
richfitz: This way is fine, but it took me a minute to work out what is going on. I think the way we do it in `orderly` is closer to 

```r
db_connect <- function(type, path) {
  x <- dettl_db_args(path, type)
  con <- do.call(DBI::dbConnect,
                 c(list(x$driver()), x$args))
  sqlite_enable_fk(con)
  con
}

sqlite_enable_fk <- function(con) {
  if (class(drv) == "SQLiteDriver") {
    ## Foreign key constraints aren't enabled by default in SQLite. Ensure
    ## they are enabled each time we connect to the db.
    DBI::dbExecute(con, "PRAGMA foreign_keys = ON")
  }
}
```

which I think is slightly easier to follow

R/default_load.R
```diff
@@ -0,0 +1,81 @@
...
+          transformed_data, table_key_pair, old_key_values, ids)
+      } else {
+        DBI::dbWriteTable(con, name, transformed_data[[name]], append = TRUE)
+      }
+    }
+  }
+}
+
+strip_primary_key_column <- function(data, primary_key) {
+  column_names <- !(names(data) == primary_key)
```
richfitz: `<- names(data) != primary_key`

?

tests/testthat/example_default_load/dettl.yml
```diff
@@ -0,0 +1,24 @@
...
+
+transform:
+  func: transform
+  test: R/test_transform.R
+
+load:
+  verification_queries: test_queries
+  test: R/test_load.R
+
+default_load: TRUE
```
richfitz: should this go into the `load` block as a key `default: true`?

richfitz: and I think the config validation needs a  check that if the default load is given then `load: func` is missing or empty

README.md
```diff
@@ -18,3 +18,7 @@ To set these create a file called `envir.yml` in the root of the directory which
 This section is only relevant if you want to run the automated package tests.
 
 Some testthat tests are configured which check the functionality of reading secrets from the vault. These tests are configured to use a test server which is controlled via `vaultr::vault_test_server()`. To set this up follow instructions in [vaultr packages vignette](https://vimc.github.io/vaultr/articles/packages.html).
+
+## Running tests which rely on PostgreSQL database
+
+Most tests run against a local SQLite database but some require a PostgreSQL database. These tests will be skipped by default when running locally. To enable them run the `scripts/test_db_start.sh` script to run up a Postgres db within a docker image. When finished testing locally run `scripts/test_db_stop.sh` to stop the image.
```
richfitz: To enable them run the `scripts/test_db_start.sh` script to run up a Postgres db within a docker ~image~ container. When finished testing locally run `scripts/test_db_stop.sh` to stop the ~image~ container.

richfitz: (an image is like a class, a container is like an instance)

richfitz (changes requested):
This looks really good.  I have one thought about the configuration that I think is worth thinking about otherwise minor things only.

I did notice a couple of things running this locally:

```
> run_import(import, c("extract", "transform", "load"))
Running extract tests /Users/rfitzjoh/Documents/Projects/epi/vimc/dettl/tests/testthat/example_default_load/R/test_extract.R
All extract tests passed.
Running transform tests /Users/rfitzjoh/Documents/Projects/epi/vimc/dettl/tests/testthat/example_default_load/R/test_transform.R
All transform tests passed.
Running load tests /Users/rfitzjoh/Documents/Projects/epi/vimc/dettl/tests/testthat/example_default_load/R/test_load.R
All tests passed, commiting changes to database.
<DataImport>
  Public:
    extract: function () 
    get_connection: function () 
    get_extracted_data: function () 
    get_transformed_data: function () 
    initialize: function (path, extract, extract_test, transform, transform_test, 
    load: function (dry_run) 
    rollback: function () 
    transform: function () 
  Private:
    con: SQLiteConnection
    extract_: function (path, con) 
    extract_test_: R/test_extract.R
    extracted_data: list
    load_: function (transformed_data, con) 
    load_test_: R/test_load.R
    path: /Users/rfitzjoh/Documents/Projects/epi/vimc/dettl/tests/ ...
    test_queries: function (con) 
    transform_: function (raw_data) 
    transform_test_: R/test_transform.R
    transformed_data: list
```

We should (IMO) print the working directory once in the log and then afterwards use relative paths only for clarity.  The `run_import` should return invisibly too to cut down on noise.  These should be done in another ticket though - would you mind?

R/db.R
r-ash: So I did it this way because the current `dbi_db_connect` method is used in several of my tests to connect to test db. I could move that connect for testing into a helper function, perhaps that will be clearer.

R/default_load.R
r-ash: Silly mistake, I've fixed that

README.md
r-ash: Thanks, fixed this too.

R/db.R
richfitz: it can always be done later if you don't think it's worth it - it's up to you

r-ash: No that's good, I've updated that now.

R/db.R
```diff
@@ -9,9 +13,22 @@ db_connect <- function(type, path) {
...
 }
 
+#' Enable foreign key constraints for SQLite connections
+#'
+#' Foreign key constraints aren't enabled by default in SQLite. Ensure
+#' they are enabled each time we connect to the db.
+#'
+#' @keywords internal
+sqlite_enable_fk <- function(con) {
+  if (class(con) == "SQLiteConnection") {
```
richfitz: Can you please update this to be `inherits(con, "SQLiteConnection")` - I should have picked that up before but that's the only safe way to ask about class membership (class vectors can be any length which makes `==` work element-wise)

R/dettl_runner.R
```diff
@@ -30,20 +36,20 @@ new_import <- function(path, db_name = NULL) {
...
 #' import completes. i.e. the load stage can be run and tests executed but the
 #' db will be rolled back.
 #'
-#' @return The processed dataImport object.
+#' @return The processed DataImport object.
 #' @export
 #'
 run_import <- function(import, run_stages = NULL, dry_run = FALSE) {
-  if (is.null(import) || class(import) != "dataImport") {
+  if (is.null(import) || class(import) != "DataImport") {
```
richfitz: as above, `class(import) != "DataImport"` should be `!inherits(import, "DataImport")`

richfitz (changes requested):
LGTM but one tiny thing to fix sorry

richfitz (approved):

PR #11, opened 2019-04-15
Add function to save extracted or transformed data as xlsx

r-ash:
This PR will
* Add `save_extracted_data` allowing users to save the extracted data from an import as an xlsx. One sheet for each item in the list
* Analogous function for transformed data `save_transformed_data`


tests/testthat/test-save-data.R
```diff
@@ -0,0 +1,83 @@
...
+
+  ## Turn off reporting when running import so import tests do not print
+  ## to avoid cluttering up test output.
+  default_reporter <- testthat::default_reporter()
+  options(testthat.default_reporter = "silent")
+  on.exit(options(testthat.default_reporter = default_reporter), add = TRUE)
+
+  import <- new_import("example/", "test")
+
+  file <- tempfile()
```
richfitz: if you add `fileext = ".xlsx"` this is much easier to test

tests/testthat/test-save-data.R
```diff
@@ -0,0 +1,83 @@
+context("save-data")
+
+test_that("extracted data can be saved", {
```
richfitz, line 3: Can you add a test with a multi-sheet export?

R/save_data.R
```diff
@@ -0,0 +1,58 @@
...
+#' current working directory called extracted_data.xlsx
+#'
+#' @export
+#'
+#' @examples
+#' import <- new_import(system.file("examples/person_information", package = "dettl"), "test")
+#' run_import(import, "extract")
+#' t <- tempfile()
+#' save_extracted_data(import, t)
+save_extracted_data <- function(import, file = "extracted_data.xlsx") {
```
richfitz: I am not sure about the file argument here.  The name is pretty arbitrary and I wonder if it would be better to either leave it off or use a temporary file if used (with same thought about the transformed version of this below)

R/save_data.R
```diff
@@ -0,0 +1,58 @@
...
+#' save_transformed_data(import, t)
+save_transformed_data <- function(import, file = "transformed_data.xlsx") {
+  save_data(import, file, "transformed")
+}
+
+save_data <- function(import, file, stage) {
+  if (!inherits(import, "dataImport")) {
+    stop(sprintf("Can't save %s data for non-DataImport object.", stage))
+  }
+  if (!file.exists(file)) {
```
richfitz: what is the role of the file create here?  the writexl package doesn't need this surely?  You might do (unconditionally) `dir.create(dirname(file), FALSE, TRUE)` if it's just the directory tree you want to ensure exists

richfitz (changes requested):
Some minor suggestions, but this will be super useful

R/save_data.R
r-ash: Yep I just want to make sure the directory exists.

R/save_data.R
r-ash: By this you think remove the default value for the arg or remove the arg altogether?

richfitz: remove the default value for the argument, keep the argument itself

richfitz (approved):

PR #13, opened 2019-04-16
Improve run_import messages

r-ash:
This PR will
* Return invisibly from `run_import` to reduce noise printed to console
* Print the path to the import directory at the start of running
* Print all other paths relative to the import directory


R/dettl.R
```diff
@@ -64,6 +64,10 @@ DataImport <- R6::R6Class(
 
     get_transformed_data = function() {
       private$transformed_data
+    },
+
+    get_import_path = function() {
```
richfitz: Nothing wrong at all with this way but if I want a read-only attribute I usually go with putting `path` into the public members and then at the end of the `initialize` function doing

```r
lockBinding("path", self)
```

which prevents it being modified.  It's a matter of taste and I know getters come very naturally from a Java perspective

R/extract_runner.R
```diff
@@ -31,3 +31,6 @@ run_extract <- function(con, extract, path, extract_test){
   }
   extracted_data
 }
+
+## work out paths as full path - working dir? If not possible then just use full path
```
richfitz: stray comment?

richfitz (changes requested):
one suggestion and one stray comment to remove

r-ash:
I'm happy to take your suggestion here and go with the `lockBinding` approach. I've removed that stray comment too.

richfitz (approved):

PR #14, opened 2019-04-29
Separate out db config code and add cfg for import log table name

r-ash:
I'm expecting [vimc-2834](https://vimc.myjetbrains.com/youtrack/issue/VIMC-2834) will have quite a few bits to it so I want to separate it out into separate PRs.

This PR will
* Make the log table name configurable
* Update `db_config.yml` to instead be called `dettl_config.yml` (now that we are also configuring vault using this file)

Couple of things I wanted to ask about here:
* Do you think after reading the config we should check that (a) a db connection can be made and (b) the log table existing in the db before proceeding? At the moment we fail when the import tries to connect to the configured data base but we could fail when we try to read the config.
* Is there anything like resource bundles in R? When writing some of the strings I think I would benefit from using this. E.g. for example updating the config name from `db_config.yml` to `dettl_config.yml` this would have been much easier if this was in a resource bundle in one location.


R/dettl_config.R
```diff
@@ -58,6 +58,12 @@ db_config_read_yaml <- function(filename, path) {
     info$db[[db_cfg]]$driver <- check_symbol_from_str(
       info$db[[db_cfg]]$driver, sprintf("%s:%s:driver", filename, db_cfg)
     )
+    if (!is.null(info$db[[db_cfg]]$log_table)) {
```
richfitz: I think it will be necessary to have the table configurable, as you have done here.

However, it's probably not something that can be *changed* once configured, or if so requires manually moving the previous table around.  Can we make sure that is appropriately documented and consider removing the possibility of a default (a default vault means that a move is more likely, a required value reduces the chance that people will want to move the table later)

R/import_config.R
```diff
@@ -0,0 +1,173 @@
+#' Get dettl config from path.
+#'
+#' Get the config containing information about location of functions containing
+#' import code.
+#'
+#' Sources any references to functions ready for use. Usea a default
```
richfitz: "Usea" => "Use"

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+#' configuration in the result that any fields are missing. This will use the
+#' field name as the expected file name and function anme. Required fields are
```
richfitz: "anme"

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+  optional <- c("rewrite_keys")
+  check_fields(info, filename, required, optional)
+  if (info$load$default) {
+    info$rewrite_keys <- ForeignKeyConstraints$new(info$rewrite_keys)
+  }
+  env <- load_sources(info$sources, path)
+  info <- read_function_fields(function_fields, info, env)
+  info$name <- basename(normalizePath(path))
+  info$path <- path
+  class(info) <- "dettl_config"
```
richfitz: You're now in a position where we have two separate objects with this class I believe (the other is created by `dettl_config.R`).  Previously you had `db_config` and `dettl_config` but now they have the same class!  It's up to you how you resolve this, but you might refactor the import to be `dettl_import_config` perhaps?

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+#'
+#' @param path Path to directory containing config.
+#'
+#' @return The config as a dettl_config object.
+#'
+#' @keywords internal
+#'
+read_config <- function(path) {
+  filename <- file.path(path, "dettl.yml")
+  assert_file_exists(path, name = "Report working directory")
```
richfitz: "Report working directory" => "Import working directory"?

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+  info <- read_function_fields(function_fields, info, env)
+  info$name <- basename(normalizePath(path))
+  info$path <- path
+  class(info) <- "dettl_config"
+  info
+}
+
+validate_load <- function(info) {
+  default <- !is.null(info$load$default) && (info$load$default || tolower(info$load$default) == "true")
+  if (xor(default, !is.null(info$load$func))) {
```
richfitz, line 47: `xor(a, !b)` is `a == b`

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+  env <- load_sources(info$sources, path)
+  info <- read_function_fields(function_fields, info, env)
+  info$name <- basename(normalizePath(path))
+  info$path <- path
+  class(info) <- "dettl_config"
+  info
+}
+
+validate_load <- function(info) {
+  default <- !is.null(info$load$default) && (info$load$default || tolower(info$load$default) == "true")
```
richfitz: `tolower(x) == "true"` looks undesirable: that looks like coercion of a logical to a string and string comparison

R/import_config.R
```diff
@@ -0,0 +1,173 @@
...
+
+set_missing_values <- function(info, field_name) {
+  ## Func can be empty for load field when running default load
+  if (field_name != "load") {
+    if (is.null(info[[field_name]]$func) || is.na(info[[field_name]]$func)) {
+      info[[field_name]]$func <- field_name
+    }
+  }
+  missing_verification_queries <- field_name == "load" &&
+    (is.null(info[[field_name]]$verification_queries) ||
```
richfitz: If this is coming from yaml is the `is.na` part needed?  If so, then pull it into a utility function (`is_missing <- function(x) is.null(x) || is.na(x)`) and use that to make this easier to read - but be aware it will only work sensibly if the value is a scalar

R/utils_assert.R
```diff
@@ -15,6 +15,14 @@ assert_scalar_character <- function(x, name = deparse(substitute(x))) {
   assert_character(x, name)
 }
 
+assert_db_name <- function(x, name = deparse(substitute(x))) {
+  assert_character(x, name)
+  if (!grepl("^[[:lower:]_]+$", x)) {
```
richfitz: numbers are allowed too I think?  So `^[[:lower:_][[:lower:]_[:digit:]]*$` (or equivalently `^[a-z_][a-z0-9_]*$`)

richfitz (changes requested):
This is a pretty noisy diff with the big move of the file over.  I've looked through by eye to convince myself that not much has really changed.

The big issue is that you have a collision in your S3 class names.  Smaller issues indicated on the code, much of which is old code sorry.

The log table does not appear to be being used - were you planning on doing that in a second PR (it seems that way in your comment).

> Do you think after reading the config we should check that (a) a db connection can be made and (b) the log table existing in the db before proceeding? At the moment we fail when the import tries to connect to the configured data base but we could fail when we try to read the config.

No, we'll fail soon enough when a connection is attempted and I think that's ok.  I can imagine cases where we'd want to load data in situations where we're off-network from the db for example

> Is there anything like resource bundles in R? When writing some of the strings I think I would benefit from using this. E.g. for example updating the config name from db_config.yml to dettl_config.yml this would have been much easier if this was in a resource bundle in one location.

This is not a concept that I am familiar with but with a quick google, I think no.  R really gives very little support for organising code in a package more than "you can put it into different files and I will concatenate them for you"

R/import_config.R
r-ash: Ah yep, overlooked that. Your suggestion is good, I've fixed that.

R/import_config.R
r-ash: Fixed now

R/import_config.R
r-ash: You're right not needed. I've removed that.

R/utils_assert.R
r-ash: Fixed this too

R/dettl_config.R
r-ash: Removed the default and I've added example and note to the vignette. Though I will flesh this out when I come to impl the work that actually does the logging.

R/import_config.R
r-ash: Fixed

R/import_config.R
r-ash: Fixed

R/import_config.R
r-ash: I'd put this in to try and make it more robust if user types "true" instead of "TRUE" Looks like yaml already handles this though and converts "true" to logical TRUE. I've removed that check.

richfitz: yaml allows all sorts of values in as True/False including yes/no and (annoyingly) y/n

R/import_config.R
r-ash: I feel using an `xor` here makes it easier to read. I want to test if either default or a load function is specified but not both or neither. Not a big preference though so I'm happy to change if you feel strongly.

richfitz: not at all, this is fine

R/import_config.R
```diff
@@ -0,0 +1,172 @@
...
+  for (field_name in fields) {
+    info <- set_missing_values(info, field_name)
+  }
+  info
+}
+
+set_missing_values <- function(info, field_name) {
+  ## Func can be empty for load field when running default load
+  if (field_name != "load") {
+    if (is.null(info[[field_name]]$func) || is.na(info[[field_name]]$func)) {
```
richfitz: these `is.na` calls can come out too, right?

R/import_config.R
```diff
@@ -0,0 +1,172 @@
...
+      info[[field_name]]$func <- field_name
+    }
+  }
+  missing_verification_queries <- field_name == "load" &&
+    is.null(info[[field_name]]$verification_queries)
+  if (missing_verification_queries) {
+    info[[field_name]]$verification_queries <- "verification_queries"
+  }
+  missing_postload_test <- field_name == "load" &&
+    (is.null(info[[field_name]]$test) || is.na(info[[field_name]]$test))
```
richfitz: `is.na` here can be removed

richfitz (changes requested):
just a couple of stray extra `is.na`'s to remove then we're good to go

R/import_config.R
r-ash: Yep, I've removed those too.

r-ash:
> The log table does not appear to be being used - were you planning on doing that in a second PR (it seems that way in your comment).

Yeah that's right. I was trying to avoid another massive PR like we had a little while ago by doing several subtasks as separate PRs and merging these into vimc-2834 before merging into master.

richfitz (approved):

PR #15, opened 2019-04-30
Check repo is up to date with remote before import can be run

r-ash:
This PR will

* Add check when running import to ensure that repo is up to date with remote. Uses `git status --porcelain`  to do the check if this returns anything it will throw an error. The import can still be run in `dry_run` mode to allow development without constantly updating git
* Add test setup to initialise git repo
* Tests now create a tempdir with git and copy files to this dir so there are changes everywhere one of the tests is actually running an import
* Add notes about `dry_run` mode to the vignette
* Updated vignette and roxygen examples to either run in `dry_run` mode or not run at all to avoid having to include setup code in the examples

One of the things I'm concerned about here is the last point. As the number of conditions for running the import goes up the test setup, vignette and roxygen examples get more and more complicated. I feel this is fine in the tests but in vignette and examples where the purpose of those is to show how to use the tool including a load of setup detracts from that. I've tried to strike a line in the middle (like running in `dry_run` mode) but I wondered if you had any thoughts on that tension?


R/dettl_runner.R
```diff
@@ -51,6 +51,9 @@ run_import <- function(import, run_stages = NULL, dry_run = FALSE) {
     stop(
       "Can only run import for non null data import with class 'DataImport'.")
   }
+  if (!dry_run && !git_is_clean(import$path)) {
+    stop("Can't run import as repository has unstaged changes. Update git or run in dry-run mode.")
```
richfitz, line 56: I would suggest adding a `force` argument here too perhaps (with default value `FALSE`) or we might find that things get a bit restrictive (i.e, the condition becomes `!force && !dry_run && !git_is_clean(import$path)`

R/dettl_runner.R
```diff
@@ -51,6 +51,9 @@ run_import <- function(import, run_stages = NULL, dry_run = FALSE) {
     stop(
       "Can only run import for non null data import with class 'DataImport'.")
   }
+  if (!dry_run && !git_is_clean(import$path)) {
```
richfitz: the git status check should probably be the root directory (i.e. with the main dettl config) not the import path itself, because that covers changes in things like the configuration itself (EDIT: though I now see that is handled by your status function, though that's not totally obvious from the call here)

R/git.R
```diff
@@ -0,0 +1,84 @@
...
+#'
+#' Check that git repo is up to date with remote by checking
+#' \code{git status --porcelian}. If import_dir is not a git root directory
+#' this will search parents of the import_dir for the git root
+#'
+#' @param import_dir The directory of the dettl import to check.
+#' @return True if repo is up to date. False otherwise.
+#'
+#' @keywords internal
+git_is_clean <- function(import_dir) {
```
richfitz: `import_dir` is is really `path`?

As indicated by the comment where this function is used, this function is probably better labelled `git_repo_is_clean` because it descends to the `.git` repo from the path given as an argument, rather than checking the status at the path that is given

R/git.R
```diff
@@ -0,0 +1,84 @@
...
+#'
+#' @param root The git project root directory.
+#' @param ignore_untracked Whether untracked files should be ignored or not.
+#' Passed as \code{--untracked-files} arg to \code{git status}
+#' @return Result from git including output and indication of success.
+#'
+#' @keywords internal
+git_status <- function(root = NULL, ignore_untracked = FALSE) {
+  args <- c("status", "--porcelain",
+            if (ignore_untracked) "--untracked-files=no")
```
richfitz: Do you have plans for this option? It's not used or tested anywhere I can see

R/git.R
```diff
@@ -0,0 +1,84 @@
...
+#' From a subdirectory within a git project locate the root directory.
+#' This is the directory contianing the .git files. Located using
+#' \code{git rev-parse --git-dir}.
+#'
+#' @param dir The directory to locate the git project root from.
+#' @return The git project root normalized, or throws an error if this can't
+#' be located.
+#'
+#' @keywords internal
+git_root_directory <- function(dir) {
```
richfitz, line 70: The other way of doing this without resorting to shelling out is to do:

```r
git_root_directory <- function(dir) {
  find_file_descend(".git", dir)
}
```

which should work reasonably well

tests/testthat/helper-git.R
```diff
@@ -0,0 +1,13 @@
+build_git_demo <- function(example_dir = "example", db_config = "db_config.yml") {
```
richfitz: Does this need changing to deal with #15 moving to `dettl_config.yml`? Why  is it necessary as an argument though? Should each example not have a configuration inside it?

richfitz (changes requested):
> Updated vignette and roxygen examples to either run in dry_run mode or not run at all to avoid having to include setup code in the examples
> One of the things I'm concerned about here is the last point. As the number of conditions for running the import goes up the test setup, vignette and roxygen examples get more and more complicated. I feel this is fine in the tests but in vignette and examples where the purpose of those is to show how to use the tool including a load of setup detracts from that. I've tried to strike a line in the middle (like running in dry_run mode) but I wondered if you had any thoughts on that tension?

I think it's not ideal to skip running the import in the vignettes, because then you can't do things like show the database working after import, and because it's ideal if the vignettes act as an integration test of the documentation.

In general, we can't modify files that are part of the package bundle if we want the package on CRAN; see [CRAN Repository Policies](https://cran.r-project.org/web/packages/policies.html)

> Packages should not write in the user’s home filespace (including clipboards), nor anywhere else on the file system apart from the R session’s temporary directory (or during installation in the location pointed to by TMPDIR: and such usage should be cleaned up). Installing into the system’s R installation (e.g., scripts to its bin directory) is not allowed.

(Unfortunately CRAN do not follow [RFC 2119](https://www.ietf.org/rfc/rfc2119.txt) and where they say "should" they really mean "must")

So to be on the safe side we'd likely have to copy the files to a temporary directory and work there anyway.  I think you do this in tests too.

What I would do in your case is to:

* write a non-exported wrapper function for preparing an example by copying it into the demo directory and setting up git (your helper-git.R file contains a function that is looking very close to this)
* use that in the vignette but in a block marked with `include = FALSE` so that the user does not see the function
* begin the discussion by describing the state of the directory

indeed, this is already what [orderly does](https://github.com/vimc/orderly/blob/65b39161e962794adfd298e418e073ffec140b90/R/testing.R#L193)

Secondarily, you might make the configuration accept configuration to make git itself optional, or a clean git optional.

R/git.R
r-ash: No plans for this, it was just copied over from orderly. Not used anywhere or tested. I just left for the case that we may want it in the future. But I guess that probably confuses things. I've taken it out.

R/git.R
r-ash: Yeah it is a path. Better naming is definitely more clear, I've updated that.

R/dettl_runner.R
r-ash: Should be more clear now function name has been updated.

R/dettl_runner.R
r-ash: That's a good idea, I've added that.

r-ash:
> Write a non-exported wrapper function for preparing an example by copying it into the demo directory and setting up git (your helper-git.R file contains a function that is looking very close to this)
> * use that in the vignette but in a block marked with `include = FALSE` so that the user does not see the function
> * begin the discussion by describing the state of the directory

I've updated this now so that everywhere we're using temp files. (So in tests, vignette and roxygen examples)

> Secondarily, you might make the configuration accept configuration to make git itself optional, or a clean git optional.

This is thinking about the use case of writing some quick import to the db? Perhaps SQL only? i.e. wanting to write some lightweight and small update the db? I think my preference would be to not add that for now and do that work in the future if such a use case comes up.

tests/testthat/helper-git.R
r-ash: I've updated the name to `dettl_config.yml`. I think we still need this arg for cases when the config is not in the current working directory. So for example in the vignette when the config is in the `inst` directory.

R/git.R
r-ash: I did think about this but thought that it would be more robust locating the directory using gits tooling. Though taking a look that is a much slower approach (however I doubt those ms are that important here). I guess shelling out is expensive!

r-ash: I'm going to update that. Just did a quick benchmark and using shell approach is 184 times slower on average.

r-ash:
I think this addresses all the points now. Apologies for another noisy diff.

tests/testthat/helper-git.R
richfitz: I see what's going on here now

richfitz (approved):

PR #16, opened 2019-05-03
Import runs are logged to the DB

r-ash:
This PR should tie together the previous two merges into vimc-2834 to make this ready for merging into master. This will:
* Log the name of the import, date it was run, optional comment, git user.name, git user.email, current active git branch and hash of HEAD to the configured log table in the database (log is done at the load stage after all other data has been imported)
* Return meaningful error if import tries to run bug log table cannot be found or does not conform to schema
* Add SQL query to `inst` directory which can be used to create the log table (this is the same as what I am planning to push to montagu-db)

For the DB date persistence we talked about in SQLite I'm using a `REAL` and in Postgres a `TIMESTAMP WITH TIME ZONE` types. In the code this will then call `Sys.time()`, convert it to UTC then save it to the DB. When we read it from Postgres using `DBI::dbGetQuery` we get a `POSIXct` type back. When reading from `SQLite` it just returns a double but we can easily convert it back. [test-log-import](https://github.com/vimc/dettl/pull/16/files#diff-25faf1ef7ad58af6d8c9863614b2b47b) checks these return the same date.


R/git.R
```diff
@@ -72,7 +74,76 @@ git_root_directory <- function(dir) {
   if (is.null(root_dir)) {
     stop(sprintf(
       "Can't run import as can't locate git directory from path %s. %s", dir,
-      "Import must be under version control to be run."))
+      "Import must be under version control to be run."
```
r-ash, line 77: Is there a good solution to multi line strings in R? This is one approach I've taken here. In `specio` I think I have created a function which strips duplicate whitespace and new lines from a string to get a similar effect. Neither approach I'm super happy with though. I'm interested in what you do?

README.md
```diff
@@ -7,6 +7,12 @@ Data extract, transform, test and load.
 
 Disinfect your workflow.
 
+## Using dettl
+
+`dettl` requires a log table to exist in the target database for an import to be run. Please see `inst/sql/create_log_table.sql` for a query to create the log table. The name of the table can be changed and configured via the import `dettl_config.yml` but the fields must remain the same.
```
richfitz: I think this should be something that the package does, TBH, even if it is done manually.

So in your verification step, you could point users at a function to initialise the table ("please run `dettl::dettl_initialise` first") (see [VIMC-2888](https://vimc.myjetbrains.com/youtrack/issue/VIMC-2888) too).

With orderly we decided that we were ok with just creating the internal table and I believe that is what flyway does too, so that would be another way to do this

R/git.R
richfitz: no, not really - I would love something like python's triple-quote strings.

I usually assemble the string in a call before the `stop`/`message` call (`msg <- ...`) and it always feels like a struggle.

Here is a recent attempt at trying to tame multi-line strings: https://github.com/vimc/orderly/blob/3434f26f4e9b57d5981b8ae47dac0e6f0e06a022/R/config.R#L43-L46

R/helper_testing.R
```diff
@@ -54,6 +56,19 @@ prepare_example_db <- function(dir, add_data = FALSE, add_job_table = FALSE) {
     DBI::dbClearResult(job_query)
   }
 
+  log_table_query <- DBI::dbSendQuery(con, sprintf(
```
richfitz: Why `dbSendQuery` + `dbClearResult` and not `dbExecute`?

R/helper_testing.R
```diff
@@ -54,6 +56,19 @@ prepare_example_db <- function(dir, add_data = FALSE, add_job_table = FALSE) {
...
+    "CREATE TABLE %s (
+      name       TEXT,
+      date       REAL,
+      comment    TEXT,
+      git_user   TEXT,
+      git_email  TEXT,
+      git_branch TEXT,
+      git_hash   TEXT
+    )",
```
richfitz: should you not be using the sql in inst?

tests/testthat/helper-db.R
```diff
@@ -19,6 +19,23 @@ prepare_example_postgres_db <- function() {
...
   )
   DBI::dbClearResult(people_query)
+
+  ## Make sure we have a fresh "log_table" table if one existed already
+  drop_log <- DBI::dbSendQuery(con,
+    "DROP TABLE IF EXISTS log_table")
+  DBI::dbClearResult(drop_log)
+  log_table_query <- DBI::dbSendQuery(con,
+    "CREATE TABLE log_table (
+      name       TEXT,
```
richfitz: the name should be the PK

R/load_runner.R
```diff
@@ -15,14 +15,18 @@
...
 #' @keywords internal
 #'
 run_load <- function(con, load, transformed_data, test_queries, path,
-                     test_file, dry_run) {
+                     test_file, dry_run, log_table, comment) {
   if (is.null(transformed_data)) {
     stop("Cannot run tests as no data has been transformed.")
   }
+  log_data <- get_log_data(path, comment)
+  verify_log_table(con, log_table, log_data)
```
richfitz, line 29: It will be worth checking here, just before load, that the import has not yet been run.

(Even if you add the name as a PK it would be nice to get a human-understandable error message here)

R/log_import.R
```diff
@@ -0,0 +1,42 @@
...
+#' @param comment An optional comment for this import run.
+#'
+#' @return The prepared data for the log table. This is the name
+#' of the import, the date, comment and git information including
+#' user name, user email, current branch and hash of HEAD.
+#'
+#' @keywords internal
+get_log_data <- function(import_path, comment) {
+  if (is.null(comment)) {
+    comment <- ""
```
richfitz: Not `NA_character_`?

R/transform_runner.R
```diff
@@ -50,24 +50,9 @@ verify_data <- function(con, transformed_data) {
...
-        stop(sprintf(
-          "Column '%s' in table '%s' returned by transform but is missing from db schema.",
-          col_name,
-          table_name
-        ))
-      }
-    }
-  }
-  ## TODO: Also check each table doesn't violate any constraints
+  lapply(names(transformed_data), function(x) {
```
richfitz: As a style thing, I'm not a big fan of `lapply` that ignores return values and would probably tend to write this as:

```
for (x in names(transformed_data)) {
 verify_table(con, x, transformed_data[[x]],
             context_info = "Transformed data")
}
```

as that makes it clearer (IMO) that we're calling `verify_table` purely for its side effects

richfitz (changes requested):
Biggest suggested change is to standardise how the creation query is run, and expose a function to help with that.

R/transform_runner.R
r-ash: That makes sense, I've updated that.

R/log_import.R
r-ash: Yep that is better than empty string. I've checked and using `NA_character` makes the db representation a `NULL` so that is better. I have updated this.

R/helper_testing.R
r-ash: Definitely, single source of truth. I've updated that. Now one query for postgres and one for sqlite in inst.

tests/testthat/helper-db.R
r-ash: Fixed that

R/helper_testing.R
r-ash: No good reason, I hadn't seen it existed. That's just what I need. I've replaced that in several places.

R/load_runner.R
r-ash: Added a check now and test that message is human readable.

README.md
r-ash: Okay fixed this now. I like the idea of an initialise method so that now exists. I've updated readme and error messages to point users to that as you suggest.

R/db.R
```diff
@@ -61,9 +76,95 @@ dettl_db_args <- function(path, type = NULL) {
...
+}
+
+#' Tests for objects of type "SQLiteConnection".
+#'
+#' @param x Object to test.
+#'
+#' @return True if object is a "SQLiteConnection".
+#'
+#' @keywords internal
+is.SQLiteConnection <- function(x) {
```
richfitz: I'm not convinced about the naming convention departure here (this is something I struggle with too, because of all the `is.*` functions, but given that `is` is generic it's a bad idea to try and follow convention).  But more of an issue is that I don't think that this is the right model (n database types, n is functions).  See comments below...

R/dettl_runner.R
```diff
@@ -42,12 +42,15 @@ dettl <- function(path, db_name = NULL) {
 #' @param dry_run If TRUE then any database changes are rolled back when the
 #' import completes. i.e. the load stage can be run and tests executed but the
 #' db will be rolled back.
+#' @param comment An optional comment to be added to the import log table for
```
richfitz, line 45: we should make sure (eventually) that these comments can be added directly to whatever script does the run.  I don't think that they should come as an extra argument to the run function necessarily

R/initialise.R
```diff
@@ -0,0 +1,45 @@
...
+#' )
+#' dettl_initialise(file.path(path, "person_information"), "test")
+#'
+dettl_initialise <- function(path, db_name) {
+  ## Get the connection info
+  path <- normalizePath(path, mustWork = TRUE)
+  con <- db_connect(db_name, path)
+  on.exit(DBI::dbDisconnect(con))
+  ## Create the table
+  if (is.SQLiteConnection(con)) {
```
richfitz: This is where the issue with the `is.*` functions becomes apparent. This might be more nicely written as a `switch` over dialect to get the appropriate path:

```
dialect <- sql_dialect(con)
path <- switch(dialect,
                sqlite = system.file("sql", "sqlite", "create_log_table.sql", package = "dettl")),
                postgres = system.file("sql", "postgres", "create_log_table.sql", package = "dettl")),
                stop(sprintf(
      "Can't initialise DB %s as not a SQLite or Postgres type. Got %s.",
       db_name, class(con)))
```

which could then be written even more easily by just passing the dialect name into the file.path call...

R/initialise.R
```diff
@@ -0,0 +1,45 @@
...
+    execute_with_return_message(con, query_text, db_name)
+  } else {
+    stop(sprintf(
+      "Can't initialise DB %s as not a SQLite or Postgres type. Got %s.",
+       db_name, class(con)))
+  }
+}
+
+execute_with_return_message <- function(con, query, db_name) {
+  tryCatch({
```
richfitz: With `tryCatch` here you will lose the traceback, - I think you want `withCallingHandlers`.

Alternatively, given what you print in both branches you might write

```r
message(sprintf("Creating log table in DB %s.", db_name))
DBI::dbExecute(con, query)
```

with no `tryCatch` as that message will provide context for the error, which is all you were catching the error for.

This function has the wrong name (the function name and the message don't line up), but can probably be removed and rolled back into the previous function as the call is duplicated there

R/log_import.R
```diff
@@ -0,0 +1,80 @@
...
+#'
+#' @param import_path Path to the import process directory.
+#' @param comment An optional comment for this import run.
+#'
+#' @return The prepared data for the log table. This is the name
+#' of the import, the date, comment and git information including
+#' user name, user email, current branch and hash of HEAD.
+#'
+#' @keywords internal
+get_log_data <- function(import_path, comment) {
```
richfitz: `build_log_data` perhaps? This name confused me because it's not getting anything as such

R/log_import.R
```diff
@@ -0,0 +1,80 @@
...
+  if (nrow(previous_runs) > 0) {
+    stop(sprintf("Import has previously been run. Previous run log:
+  name:           %s
+  date:           %s
+  comment:        %s
+  git user.name:  %s
+  git user.email: %s
+  git branch:     %s
+  git hash:       %s",
+                 previous_runs[1, ]$name,
```
richfitz: small thing, but prefer `previous_runs$name[[1]]` here as that will be more predictable with types

richfitz (changes requested):
OK, smaller things in here this time.

One thing that I think is worth thinking about is the position of the comment - I think it might be more at home within an individual import rather than relying on an explicit argument (that will be needed by the time we want to do, say, "run all imports that are pending").

The other is the dialect treatment and I am more than happy to discuss this in person if that would be useful

R/log_import.R
r-ash: Yep that's definitely more accurate for what it is doing. I just can't resist writing getters :)

I've made that change.

R/log_import.R
r-ash: That's super neat, I was doing that assuming that if there was ever 2 rows returned for some reason it would work around it and not throw an error. Just seen if `sprintf` gets called with vector values it prints multiple lines which is very nice.

R/dettl_runner.R
r-ash: I think I will create a separate ticket for this as this PR is already pretty big.

https://vimc.myjetbrains.com/youtrack/issue/VIMC-2908

R/initialise.R
r-ash: Yeah that's nice, I do just want it for context. I will just print the message beforehand.

R/initialise.R
r-ash: That is nicer, I've tidied this up now using a `sql_dialect` function.

R/log_import.R
```diff
@@ -0,0 +1,80 @@
...
+  if (nrow(previous_runs) > 0) {
+    stop(sprintf("Import has previously been run. Previous run log:
+  name:           %s
+  date:           %s
+  comment:        %s
+  git user.name:  %s
+  git user.email: %s
+  git branch:     %s
+  git hash:       %s",
+                 previous_runs[[1]],
```
richfitz: sorry, this should be `previous_runs$name`, etc - i.e., refer to the variable by name not position.  I'm looking at the previous comment https://github.com/vimc/dettl/pull/16#discussion_r282474354 and can see the confusion.  But we're never going to get >1 run here because the name is the PK

richfitz (changes requested):

richfitz (approved):

PR #18, opened 2019-05-21
VIMC 2837 - Improve public API

r-ash:
This PR will
* Make function names more consistent
* Remove `run_import` in favour of working with the object itself

In a separate PR I will
* Add documentation for R6 classes
* Add new functions for running the import (without having to create an import object first)


R/initialise.R
```diff
@@ -11,9 +11,9 @@
 #'   system.file("examples", "dettl_config.yml", package = "dettl"),
 #'   add_log_table = FALSE
 #' )
-#' dettl_initialise(file.path(path, "person_information"), "test")
+#' dettl_db_create_log_table(file.path(path, "person_information"), "test")
```
richfitz: what do you think about `dettl::` prefixes to all examples?  They run with the package attached so it's not strictly necessary but it seems that's how we'd tend to use the package.  Should be followed through the vignettes (which currently do use `dettl::`) and (eventually) in any examples that end up in the README

vignettes/dettl.Rmd
```diff
@@ -111,11 +111,11 @@ The process at a minimum needs code to define 3 functions one for each stage of
 
 ## Creating a new ETL process
 
-If you want to create a new process within an existing project this can be done by opening the project as an RStudio project and using `new_dettl()`. Note that the use of `withr::with_dir(tempdir(), ...)` in the example below is because this is running in a vignette and the code is an example. This is not necessary when using `dettl` in a real environment.
+If you want to create a new process within an existing project this can be done by opening the project as an RStudio project and using `dettl_new()`. Note that the use of `withr::with_dir(tempdir(), ...)` in the example below is because this is running in a vignette and the code is an example. This is not necessary when using `dettl` in a real environment.
```
richfitz: Not strictly related to this PR but the "new process" here is confusing me because I read [instance of a computer program](https://en.wikipedia.org/wiki/Process_(computing))

R/new.R
```diff
@@ -7,8 +7,8 @@
 #' @export
 #'
 #' @examples
-#' new_dettl("test import")
-new_dettl <- function(name) {
+#' dettl_new("test import")
+dettl_new <- function(name) {
```
richfitz, line 11: If you took a `path` argument here too then the examples would be easier.  Can push this into a new PR if you would rather

vignettes/dettl.Rmd
```diff
@@ -111,11 +111,11 @@ The process at a minimum needs code to define 3 functions one for each stage of
...
 
 ```{r}
 withr::with_dir(tempdir(), 
-  import_name <- dettl::new_dettl("person_information")
+  import_name <- dettl::dettl_new("person_information")
```
richfitz: See the comment above about a `path` argument.  I also wonder how much sense it makes to be able to run this in a directory that does not contain a `dettl_config.yml` file (another separate ticket certainly, if you agree)

richfitz: I also prefer the form

```r
import_name <- withr::with_dir(tempdir(), 
  dettl::dettl_new("person_information")
)
```

because it's less ambiguous about scoping rules - I'd missed the assignment here on the first read

R/save_data.R
```diff
@@ -13,51 +15,44 @@
...
 
   if (is.null(data)) {
-    stop(sprintf("Can't save %s data as stage has not been run.", stage))
+    stop(sprintf("Can't save %s data as stage has not been run.",
+         paste(stage, collapse = ", ")))
   }
   writexl::write_xlsx(data, file)
-  message(sprintf("Saved %s data to %s", stage, file))
+  messages <- sprintf("Saved %s data to %s", stage, file)
+  for (message in messages) {
```
richfitz: or
```
message(paste(messages, collapse = "\n"))
```

richfitz (changes requested):
minor things here, mostly in the form of other tickets to create.  otherwise this looks like a big improvement

vignettes/dettl.Rmd
r-ash: Yes that is a bit confusing. And I think talking about it as a "process" is not really a thing that occurs anywhere other than in the vignette. I've updated that in favour of talking about an import.

R/new.R
r-ash: New ticket for that VIMC-2927

vignettes/dettl.Rmd
r-ash: I've added a PR for first point. There would be little reason to do `dettl_new` unless your working directory contained a config file.

I've updated the code re the second point, I didn't know you could drive `withr` like that.

R/save_data.R
r-ash: That's nicer as we can avoid the for loop. I've updated that.

r-ash:
I've done the work here on getting the code coverage to 100%. I think mostly this has been straightforward. For the coverage of `utils_assert.R` I've had to do some pretty heavy handed mocking to imitate windows file system so we can see how it would behave if file system.

R/load_runner.R
```diff
@@ -30,8 +30,6 @@ run_load <- function(con, load, transformed_data, test_queries, path,
   verify_first_run(con, log_table, log_data)
   before <- test_queries(con)
   DBI::dbBegin(con)
-  transaction_active <- TRUE
-  on.exit(if (transaction_active) {DBI::dbRollback(con)})
```
r-ash, line 34: I also wanted to highlight this change as I removed it as it wasn't being hit by the code coverage.

So I had added this a bit ago to handle the case what happens if there is some unexpected error in the code causing it to stop before the transaction is committed or rolledback. Looking at the documentation for DBI it looks like this is not necessary.

> Disconnection from a connection with an open transaction effectively rolls back the transaction. All data written in such a transaction must be removed after the transaction is rolled back.

richfitz (approved):

PR #21, opened 2019-06-19
Read foreign key constraints from the database for use in default load

r-ash:
This PR will
* Add functions to get foreign key constraints from postgres and sqlite databases
* Remove configuration for setting up foreign key constraints
* Use automatic reading of foreign key constraints in default load instead of configured constraints


NEWS.md
```diff
@@ -0,0 +1,3 @@
+## 0.0.3
+
+Read foreign key constraints automatically from the database
```
richfitz: It would be nice to add a (VIMC-2859) here I think so we can relate features back to the original ticket

R/foreign_keys.R
```diff
@@ -0,0 +1,89 @@
...
+#'
+#' @return A list representing foreign key constraints
+#' @keywords internal
+get_fk_constraints <- function(con) {
+  dialect <- sql_dialect(con)
+  constraints <- switch(
+    dialect,
+    "sqlite" = get_sqlite_fk(con),
+    "postgresql" = get_postgres_fk(con, dialect),
+    {
```
richfitz: you don't need curlys here for this (and it's more typical not to use them)

R/foreign_keys.R
```diff
@@ -0,0 +1,89 @@
...
+#'
+#' @param con Connection to get constraints for.
+#'
+#' @return A list representing foreign key constraints
+#' @keywords internal
+get_fk_constraints <- function(con) {
+  dialect <- sql_dialect(con)
+  constraints <- switch(
+    dialect,
+    "sqlite" = get_sqlite_fk(con),
```
richfitz: If you drop the dialect argument from `get_postgres_fk` (as it's a constant in this case) you can write:

```
get_fk <- switch(dialect,
                 postgresql = get_postgres_fk
                 sqlite = get_postgres_fk,
                 stop(sprintf(...)))
constraints <- get_fk(con)
```

R/foreign_keys.R
```diff
@@ -0,0 +1,89 @@
...
+#' Get FK constraints for SQLite connection
+#'
+#' @param con SQLite connection to get constraints for.
+#'
+#' @return A data frame representing foreign key constraints.
+#' @keywords internal
+get_sqlite_fk <- function(con) {
+  tables <- DBI::dbListTables(con)
+
+  constraints <- data.frame(matrix(ncol = 4, nrow = 0),
```
richfitz:
```
data_frame(constraint_table = character(0), constraint_column = character(0), ...)
```

etc?

R/foreign_keys.R
```diff
@@ -0,0 +1,91 @@
...
+}
+
+#' Get FK constraints for postgres connection
+#'
+#' @param con Postgres connection to get constraints for.
+#'
+#' @return A data frame representing foreign key constraints.
+#' @keywords internal
+get_postgres_fk <- function(con) {
+  path <- system.file("sql", "postgresql", "get_fk.sql", package = "dettl")
```
richfitz: Just a suggestion, but if you're getting a lot of files from the package, you might want to write a helper function:

```r
dettl_file <- function(...) {
  system.file(..., package = "dettl", mustWork = TRUE)
}
```

which you'd use above as:

```r
dettl_file("sql", "postgresql", "get_fk.sql")
```

(the `mustWork` argument errors if not found, which catches a number of bugs)

inst/sql/postgresql/get_fk.sql
```diff
@@ -0,0 +1,24 @@
+WITH unnested_confkey AS (
```
richfitz, line 1: nice job working all this out :hand:

R/foreign_keys.R
```diff
@@ -0,0 +1,91 @@
...
+        pragma.'to' as referenced_column,
+        '%s' as constraint_table
+        FROM  pragma_foreign_key_list('%s') as pragma",
+      table, table)
+  })
+
+  constraints <- lapply(queries, function(query) {
+    DBI::dbGetQuery(con, query)
+  })
+  as.data.frame(do.call(rbind, constraints))
```
richfitz: is the `as.data.frame` needed or wanted here? Each of the result of `dbGetQuery` will be a df so rbinding these together will give you a df

R/helper_testing.R
```diff
@@ -60,10 +60,66 @@ prepare_example_db <- function(dir, add_data = FALSE, add_job_table = FALSE,
...
+#'
+#' @keywords internal
+add_fk_data <- function(con) {
+  DBI::dbExecute(con,
+    "CREATE TABLE region (
+      name TEXT PRIMARY KEY,
+      parent TEXT,
+      FOREIGN KEY (parent) REFERENCES region(name)
+    )")
+    region1 <- data.frame(list(
```
richfitz: what is the `list()` call doing here? (& other `data.frame(list(...))` constructions throughout)

tests/testthat/example_default_load/dettl.yml
```diff
@@ -15,9 +15,3 @@ load:
   default: TRUE
```
richfitz, line 15: Not necessarily for this PR, but I wonder if "automatic" is better than "default" as a name for this

richfitz (commented):
Super minor things in here - this looks really good and I'm super keen to see how it goes in practice.

I think that this would warrant it's own separate blog post and/or vignette once we get an example - this has always felt like a real pain point with SQL and I think people will be really interested in it - **but** we need a more obvious way of getting it across, as was apparent at standup

richfitz (changes requested):
Nice job on this - super minor comments only here.

I think that this would warrant it's own separate blog post and/or vignette once we get an example - this has always felt like a real pain point with SQL and I think people will be really interested in it - **but** we need a more obvious way of getting it across, as was apparent at standup

R/foreign_keys.R
richfitz: (this was left over from our discussion before)

R/foreign_keys.R
r-ash: Much nicer!

R/foreign_keys.R
r-ash: Got rid of those.

NEWS.md
r-ash: I've added that

R/foreign_keys.R
r-ash: That will definitely tidy stuff up a little, I've added that.

R/foreign_keys.R
r-ash: Not needed, I didn't try to see what would happen without the cast. I've removed it now.

R/helper_testing.R
r-ash: Not necessary again, I've removed those too.

tests/testthat/example_default_load/dettl.yml
r-ash: Yeah agreed I think that is a better name for what it is doing

R/dettl.R
```diff
@@ -15,8 +15,8 @@
 ##'
 ##' @examples
 ##' path <- dettl:::prepare_test_import(
-##'   system.file("examples", "person_information", package = "dettl"),
-##'   system.file("examples", "dettl_config.yml", package = "dettl"))
+##'   dettl:::dettl_file("examples", "person_information"),
```
richfitz: The examples should continue to use `system.file` - they should use only public API features

richfitz (changes requested):
One tiny thing sorry, then we'll merge this

R/dettl.R
r-ash: I see, I should be exporting `prepare_test_import` then too?

r-ash: I've updated that back to using `system.file`. I've also gone through the examples and made sure I'm only using exported functions.

richfitz (approved):

PR #23, opened 2019-06-26
Automatic upload should only strip autogenerated fk constraints

r-ash:
Apologies this one has kind of grown. The big changes here are there is a big change in how `insert_into_returning` works so that it can return more than one column. And we add code to detect whether a column is a `serial` column and then use that in the default/automatic load.

This PR will
* Handle the case when we have more than 1 referenced column with a single table more gracefully
   * This includes updating the IR of constraints to support multiple
   * Updating `insert_into_returning` to be able to return multiple columns
   * Removing references to `primary_key` within the documentation and code (was assuming any referenced key was a primary key previously which is not the case)
* Detect `serial` columns in postgres and sqlite
* Only do the row stripping and id replacing steps during automatic upload for the `serial` columns, for non-serial columns just tries to append the columns as they are in the data frame
* Wrap the db update call within a `withCallingHandlers` to add some context info should db update fail (e.g. if violates a primary key constraint)
* Add some notes about other edge cases, in particular what if we have 1 column which has a fk constraint on another column within the same table (at the moment this will never update on upload as we try and append the whole data frame of new rows into db in one go) to support that would require moving to appending 1 row at a time.


R/default_load.R
```diff
@@ -18,43 +18,74 @@ get_default_load <- function() {
...
+          transformed_data <- update_child_tables(
+            transformed_data, table_key_pair, old_key_values, return[[col]], name)
+        }
       } else {
-        DBI::dbWriteTable(con, name, transformed_data[[name]], append = TRUE)
+        withCallingHandlers(
+          DBI::dbWriteTable(con, name, transformed_data[[name]], append = TRUE),
+          error = function(e) {
+            readable_data <- paste(
+              utils::capture.output(print(transformed_data[[name]])),
```
r-ash: I wanted to ask about this bit. This is a suggestion from a SO post for how to format a set of data for debugging. I can see some problems here for if we have some huge data set. I wonder how best to handle this. I want to return something for some context, perhaps output of `str()` would be better? What do you think?

richfitz: Make a classed error and set the data to be part of the error, then write a print method for it.  The nice thing about that is then you can catch the error and get the actual data

R/insert_into_returning.R
```diff
@@ -8,88 +8,181 @@
...
-        rowid_last_entry <- DBI::dbGetQuery(con, "SELECT last_insert_rowid()")[1, 1]
-        ret <- DBI::dbGetQuery(con, sprintf("SELECT %s FROM %s WHERE rowid = $1", id, table), rowid_last_entry)[[1]]
-      }
-      ret
+      ret <- sqlite_insert_empty_row(con, table, return)
     }
+    ret
   }
 
   if (!is.data.frame(d)) {
```
richfitz: As discussed, I think this should come out

R/insert_into_returning.R
```diff
@@ -8,88 +8,181 @@
...
     d <- as.data.frame(d, stringsAsFactors = FALSE)
   }
-  lapply(seq_len(nrow(d)), insert1)
+  rows <- lapply(seq_len(nrow(d)), insert1)
+  do.call(rbind.data.frame, rows)
+}
+
+sqlite_insert_row <- function(con, table, x, key, return) {
```
richfitz: tiny thing, but avoid `return` here - `ret`, `result`, `res` etc all fine (and elsewhere here)

R/utils.R
```diff
@@ -253,3 +253,5 @@ zip_dir <- function(path, dest = paste0(basename(path), ".zip")) {
 dettl_file <- function(...) {
   system.file(..., package = "dettl", mustWork = TRUE)
 }
+
+data_frame <- function(...) { data.frame(..., stringsAsFactors = FALSE) }
```
richfitz: Formatting

R/insert_into_returning.R
```diff
@@ -8,88 +8,181 @@
...
-             "VALUES",
-             sprintf("  (%s)", paste0("$", seq_along(x), collapse = ", "))
-            )
-    sql <- paste(sql, collapse = "\n")
-    if (is.null(key)) {
-      DBI::dbExecute(con, sql, unname(x))
-      ## TODO: What happens here if we insert but the table has no rowid?
-      rowid_last_entry <- DBI::dbGetQuery(con, "SELECT last_insert_rowid()")[1, 1]
-      DBI::dbGetQuery(con, sprintf("SELECT %s FROM %s WHERE rowid = $2", id, table), rowid_last_entry)[[1]]
+    if (length((names(x))) > 0) {
```
richfitz: `length((names(x)))` should be `ncol(x)` I think, and that `as.list` can go away (otherwise just `length(x)`)

richfitz (changes requested):
honestly, given the size of this it looks reasonable - you've got way more in tests than code changes.

The main thing I'd really like to see is a little human readable bit of pseudocode above the implementation details for the "insert into" bits, as there's clearly a lot of logic there in your head that is not clear to me.  It's possible that it will be no clearer in prose than in code though

R/default_load.R
r-ash: Added a custom error now `data_write_error.R`. It doesn't have a specific print method - but we do build up a message which is printed when the error is thrown.

We need to add a print method too for if we want to catch this downstream?

r-ash:
I think this addresses the raised points. I've added notes to `insert_into_returning` which I hope explain the intention of why it is doing certain things. Not put some pseudocode above, let me know if intention is still not clear and I can do that. 

Also added a custom error class but this doesn't have a print method. It does set up a message though. We did discuss why we may need a print method. Could you remind me again?

richfitz (approved):

PR #24, opened 2019-07-04
Vimc 3004

weshinsley:
I think this is done now - passes on travis after a bit of hacking.

One test I've commented - head detachment - see test-git.R:61
I don't know how to do that with gert - my attempt just treats "HEAD~1" as a literal branch name.

Issue: https://vimc.myjetbrains.com/youtrack/issue/VIMC-3004

NB also - a couple of tests fail in annoying ways on windows because of filenames with/without trailing forward/back slashes in different places - there are slightly different "interpretations" between OSes - I'll return to that later perhaps when I have more time - they aren't important failures.


R/git.R
```diff
@@ -102,14 +56,21 @@ git_user <- function(path) {
...
-  args <- c("config", field)
-  res <- git_run(args, root = path, check = TRUE)
-  res$output
+  gitres <- gert::git_config(repo = git_root_directory(path))
+  gitres <- rbind(
+    gitres[gitres$level == 'local', ],
+    gitres[gitres$level != 'local', ])
+
+  if (!any(gitres$name == field)) {
+    stop("git config field not found")
```
r-ash: Could we return the `field` and `path` they asked for here? Thinking about if this is thrown when we're several layers deep in the call stack - more context is always useful. Or if we go and call this in future from somewhere where the calling method has a less clear name.

.travis.yml
```diff
@@ -5,6 +5,7 @@ sudo: false
 cache: packages
 r_packages:
   - covr
+  - gert
```
r-ash: Just curious - why is `gert` required by travis?

tests/testthat/test-git.R
```diff
@@ -11,30 +11,30 @@ test_that("git parent directory can be located", {
...
+    "Import must be under version control to be run."), fixed = TRUE)
 
 })
 
 test_that("git status can be retrieved and clean can be checked", {
   path <- build_git_demo()
-  expect_equal(git_status(path),
-               list(success = TRUE, code = 0,
-                    output = character(0), clean = TRUE))
+#  expect_equal(git_status(path),
```
r-ash: I think we can remove the tests of `git_status`. This is no longer a function we implement (or want to implement in the future) so I don't see much reason to keep them around.

tests/testthat/test-git.R
```diff
@@ -11,30 +11,30 @@ test_that("git parent directory can be located", {
...
   expect_false(git_repo_is_clean(path))
 })
 
-test_that("git run returns errors", {
-  path <- build_git_demo()
-  writeLines("hello", file.path(path, "hello"))
-  expect_error(git_run(c("test"), path, TRUE),
-               "Error code 1 running command:[.\n]+")
-})
+#test_that("git run returns errors", {
```
r-ash: Same here, now we no longer impl `git_run` I think we can remove this test

R/utils.R
```diff
@@ -223,22 +223,22 @@ file_copy <- function(..., overwrite = TRUE) {
...
-
-system3 <- function(command, args) {
-  res <- suppressWarnings(system2(command, args, stdout = TRUE, stderr = TRUE))
-  code <- attr(res, "status") %||% 0
-  attr(res, "status") <- NULL
-  list(success = code == 0,
-       code = code,
-       output = res)
-}
+#sys_which <- function(name) {
```
r-ash: I think we can remove these lines too if they're no longer needed.

tests/testthat/test-git.R
```diff
@@ -44,20 +44,31 @@ test_that("git user and email can be retrieved", {
...
+  expect_error(git_config(path, "favourite_vegetable"),
+               "git config field not found")
+})
+
 test_that("git branch can be retrieved", {
   path <- build_git_demo()
   branch <- git_branch(path)
   expect_equal(branch, "master")
 
+# Not sure how to detach head with gert, so will
```
r-ash: I think `git_info` has the information we need to work this out. Doing an info with a detached head we get

```
$path
[1] "/home/rashton/projects/dettl/"

$head
[1] "HEAD"

$shorthand
[1] "HEAD"

$commit
[1] "724dc885afe37a79caeb64bde3a742911a317f07"

$remote
[1] NA

$upstream
[1] NA

...
```
Whereas when we're on a branch we have 
```
$head
[1] "refs/heads/vimc-3004"
...
```
I think probably enough to check that `$head` points to an actual branch i.e. `git_info()$head %in% git_info()$reflist`

What do you think?

r-ash (changes requested):

R/git.R
weshinsley: Done

tests/testthat/test-git.R
weshinsley: Great - thought I'd comment them and give you a look before I obliterated them. Done

tests/testthat/test-git.R
weshinsley: Done

R/utils.R
weshinsley: Done

.travis.yml
weshinsley: I misunderstood .travis.yml when I was trying to get libgit2 installed! Removed...

tests/testthat/test-git.R
weshinsley: Fixed with mockery!

r-ash (approved):

r-ash:
Great stuff, thanks for this.

PR #25, opened 2019-07-09
Fix transaction not being rolled back when db import fails

r-ash:
This PR will

* Rollback transaction started during load should any error be thrown from load

This should fix the problems you were seeing where you needed to manually rollback the connection to run the `load` step again.

### DIFF
--- a/R/load_runner.R
+++ b/R/load_runner.R
@@ -28,8 +28,21 @@ run_load <- function(con, load, transformed_data, test_queries, path,
   log_data <- build_log_data(path, comment)
   verify_log_table(con, log_table, log_data)
   verify_first_run(con, log_table, log_data)
-  before <- test_queries(con)
   DBI::dbBegin(con)
+  withCallingHandlers(
+    do_load(con, load, transformed_data, path, test_file, test_queries,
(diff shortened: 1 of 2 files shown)


weshinsley (changes requested):
Not sure if this is working yet - when I test (with some error or other in my data...)
```
> con <- import$get_connection()
> import$load(dry_run=TRUE)
Running load C:/Files/Dev/R/montagu-imports/2019-07-01_i2938_201906who_measles_rubella_elim
Error in vapply(X, FUN, numeric(1), ...) : values must be length 1,
 but FUN(X[[1]]) result is length 0
> DBI::dbRollback(con)
> DBI::dbRollback(con)
Error: Call dbBegin() to start a transaction.
```
- I think that trace means the transaction wasn't rolled back at first, or I'd get the error first time? I'll try and emulate the actual postgres error we got and see if that behaves differently to before...
- (Prior to this, I did ` devtools::install_github("https://github.com/vimc/dettl", ref = "vimc-3004")` which hopefully means I'm testing against the right code?)

This is all rubbish, it works.

weshinsley (approved):

PR #26, opened 2019-07-09
Expose automatic load function

r-ash:
This PR will

* Replace references to default load to automatic load throughout
* Exposes `dettl_auto_load` so it can be called from a custom load function (e.g. for if we want to upload data and then do an update to fix any special cases like we have with a cyclic fks with `model_version` and `model` tables)

Sorry this looks like a big one, most of the changes are just text changes.


weshinsley (approved):
Looks great!

PR #27, opened 2019-07-11
Nullable FK constraints are supported

r-ash:
Apologies this should be on branch vimc-3014. We've merged some changes on -3015 previously as in terms of using the incorrect name this shouldn't be a huge problem. I've added a note on issue -3014 referencing this.

This PR will
* Add ability to read not null constraints from the DB
* Add check at end of transform stage (if table is missing column which are non-serial and are non-nullable then fail with a useful message)
* Update mapping of old values to new values on automatic upload to support having NAs within the column (or referenced keys to values existing in the db before this import was run) this means we can now do an import where a fk references newly added and previously existing values


weshinsley (approved):
Looks good Rob, I've tested on my code, and I can now add NAs to a foreign key.
I haven't done a super-detailed code review - but I think at a glance it looks good to merge in and get on with this import...!

PR #28, opened 2019-07-11
Fix bug where code expects all referenced keys to exist within the import

r-ash:
This PR will

* Allow upload of data without specifying autogenerated columns which are referenced by some fk constraint when no other uploaded data references that column
* Throw a useful error should the autogenerated column not be specified when it is referenced by another table within the same upload


weshinsley (approved):
Really useful change, tested and working well.

PR #29, opened 2019-07-12
Running import on prod requires user confirmation

r-ash:
This PR will

* Allow users to configure whether an import needs to be confirmed (via a yes no question) before `load` stage is run


weshinsley (approved):
Tested with an import on science (even with dry_run = TRUE), and got a nice warning message.
All looks good to me

PR #31, opened 2019-08-29
Add extracted_data to test environments

weshinsley:
So this PR adds extracted_data into the environment for testing the transform, and adds both the exctracted_data and transformed_data into the environment for testing the load.

For an example that uses this, see https://github.com/vimc/montagu-imports/pull/14 , where the number of rows changed (ie, test_queries and the load tests), and the number of rows in the transformed data depend on the CSV file that I load in at the meta stage. It's intended to be an easily reusable import, should we want to do more adding of burden outcomes to expectations. 

Strictly speaking, I didn't need the transformed_data in my test_load, but it seemed odd to only add the extracted_data and not transformed_data. I suppose one could write more tests that ask, "did all the transformed data *really* end up in this database table, or something...

Anyway, see what you think, and whether there's any better way to fix this in dettl, or in my montagu-import PR...

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dettl
 Title: Data extract, transform, test and load
-Version: 0.0.7
+Version: 0.0.8
 Authors@R: 
     person(given = "Robert",
            family = "Ashton",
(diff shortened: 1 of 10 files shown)


r-ash (approved):
Looks good to me.

I see no reason why we can't make the `extracted_data` and `transformed_data` available to all test environments. 

I can definitely see the use case for this too - especially as we're moving towards more reusable imports.

PR #32, opened 2019-10-03
Allow re-running of same import

r-ash:
This PR will
- Allow users to re-run an import which has already been run using `force = TRUE`
   - This will skip checking that the import has not been run before and will instead append `_n` to the name in the import log where n is the number of times the import has already been run. This is super hacky and should be removed when we properly support re-running imports.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,6 +1,6 @@
 Package: dettl
 Title: Data extract, transform, test and load
-Version: 0.0.8
+Version: 0.0.9
 Authors@R: 
     person(given = "Robert",
            family = "Ashton",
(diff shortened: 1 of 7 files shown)


weshinsley (commented):
The PR looks great in itself Rob - but when time allows, @richfitz will review whether this is the right approach - (ie whether I've requested something non-ideal for the bigger picture...)

r-ash:
> The PR looks great in itself Rob - but when time allows, @richfitz will review whether this is the right approach - (ie whether I've requested something non-ideal for the bigger picture...)

I definitely think this isn't the right approach - it is more a thing we can use temporarily until we can spend some more time on implementing a proper approach

richfitz (changes requested):
Even as a hack, I don't think that overloading the `force` argument here is a good idea as now it will do two very different things (allow running with unclean git and allow running a second time). Can we get a different name - perhaps on both arguments (`allow_unclean_git`, `allow_rerun` perhaps, though these names are also a bit rubbish)

PR #33, opened 2019-11-06
VIMC-3035: reload dettl sources

richfitz:
This PR adds the ability to reload dettl sources without recreating the dettl object. The use case I had for this was debugging a malfunctioning `load` step, while not wanting to rerun a long-running `extract` step.

This partly fixes VIMC-3154 too because the reload will refresh the db connection, which means that a failed transaction is not fatal.

Note that in its current form, the extracted and transformed data remain in the object (probably wanted?).


r-ash (approved):
Looks good

PR #34, opened 2019-11-12
VIMC-3217: option to require a particular branch to run

richfitz:
This PR will add a new configuration option that specifies a branch required for running a report.  For example, in montagu-imports, we should set this to `master` for the `production` database.

This is only going to stop accidental running of reports on production but would still be useful.


r-ash (approved):
Nice

PR #35, opened 2020-02-11
Add mode to dettl config

r-ash:
This PR will

* Add a mode option to dettl config
   * By default mode is "append" if mode is set to "create" then we can only create new tables and not append to existing ones


R/dettl.R
```diff
@@ -58,8 +59,14 @@ DataImport <- R6::R6Class(
 
     reload = function() {
       dettl_config <- read_config(self$path)
+      mode <- dettl_config$dettl$mode
```
richfitz: This seems like more work than needed - can you not just pass the mode directly though your check function and have it return a validated one, so that this becomes

```
private$mode <- check_valid_mode(private$mode)
```

(or pass in `private$mode %||% "append"`) - I don't think you want to allow NA in here and in any case, that test will do unwanted things with `mode = c(NA, 1)` and `mode = c(1, NA)`.  In orderly we have a utility `match_value` that you might copy across for this bit of drudgery

R/mode.R
```diff
@@ -0,0 +1,24 @@
...
+  valid_modes <- c("append", "create")
+  if (!(mode %in% valid_modes)) {
+    stop(sprintf('Invalid mode - mode must be one of %s got "%s".',
+                 paste(valid_modes, collapse = ", "), mode))
+  }
+  invisible(TRUE)
+}
+
+allow_create_table <- function(mode) {
+  allow <- FALSE
```
richfitz: Anything wrong with the body here being `mode == "create"`

R/mode.R
```diff
@@ -0,0 +1,24 @@
...
+  if (mode == "create") {
+    allow <- TRUE
+  }
+  allow
+}
+
+get_auto_load_function <- function(mode) {
+  auto_load_function <- dettl_auto_load
```
richfitz: consider

```
switch(mode,
       create = dettl_auto_load_create,
       append = dettl_auto_load)
```

richfitz (changes requested):
This is super tidy - nice one.  Can you add the mode to the template in inst?  Other suggestions are just suggestions - take or leave as you want

R/dettl.R
r-ash: Thanks - I have tidied that up.

R/mode.R
r-ash: Nope - that's much nicer.

R/mode.R
r-ash: Changed that

R/dettl.R
```diff
@@ -134,6 +137,10 @@ DataImport <- R6::R6Class(
       invisible(TRUE)
     },
 
+    get_mode = function() {
```
richfitz: This appears untested and/or unused - could you test or remove?

richfitz (changes requested):
sorry, one last thing

R/dettl.R
r-ash: Crap added that when I was debugging a thing - forgot to remove it. Done that now - thanks.

richfitz (approved):

PR #36, opened 2020-02-11
Fix bug where import could only be run if import object was created from root of import dir

r-ash:
This PR will
* Remove `path` arg from `extract` function - this will break backwards compatibility
* Instead set the working dir as the dir the import object was created with - this is the root of the import being done - so all paths should be relative to this (which is how it has been written previously) 

This fixes the bug where the import can only be run if you create the object from the root of the import directory. The removed path arg I have not seen any instances of being used - everyone was writing paths relative to the root of the import dir.


inst/examples/person_information/R/extract.R
```diff
@@ -3,14 +3,13 @@
...
 #' it into memory ready for transform stage.
 #'
-#' @param path Path to the import project root used for finding any local data.
 #' @param con The active DBI connection for extracting any data.
 #'
 #' @return A list of data frames representing the extracted data.
 #'
 #' @keywords internal
-extract <- function(path, con) {
+extract <- function(con) {
```
richfitz, line 11: This is definitely nicer.

inst/examples/person_information/R/extract.R
```diff
@@ -3,14 +3,13 @@
...
   raw_data <- list()
-  raw_data$people <- read.csv(file.path(path, "data/people.csv"), stringsAsFactors = FALSE)
+  raw_data$people <- read.csv(file.path("data", "people.csv"), stringsAsFactors = FALSE)
```
richfitz: This should just be

```
read.csv("data/people.csv", stringsAsFactors = FALSE)
```

as I was initially confused as to why `file.path` was needed here - and through other affected lines in this PR incl. the vignette

richfitz (changes requested):
This is much nicer. If you could tidy up the `file.path` usage in examples this is good to go

inst/examples/person_information/R/extract.R
r-ash: This might be some confusion on my end. I thought we still needed `file.path` here to fix file separator for a particular platform. Windows uses `\`? So hard coding the path as `data/people.csv` wouldn't work for windows users? Is that true or does R work around that some other way?

richfitz: Both `a\\b` and `a/b` work on windows from R (as you can see in the original code that used `data/people.csv` as the second arg to `file.path`.  In fact, the help for `?file.path` says

> The components are by default separated by ‘/’ (not ‘\’) on
> Windows.

We should probably put some appveyor tests on this package in any case at some point (not in this PR!) though

richfitz (approved):

PR #37, opened 2020-02-12
vimc-3432: Add basic verbose output to automatic load

richfitz:
Small usability tweak as some of the uploads take a while. Could be obviously reformatted a lot more nicely...


r-ash (approved):
Looks good

PR #38, opened 2020-02-13
Add pre and post load sections

r-ash:
This PR will
* Allow users to specify pre and post load functions


R/dettl.R
```diff
@@ -74,6 +76,8 @@ DataImport <- R6::R6Class(
       private$transform_ <- dettl_config$transform$func
       private$transform_test_ <- dettl_config$transform$test
       private$load_ <- load_func
+      private$pre_load_ <- dettl_config$load$pre
```
richfitz: `load_pre_` and `load_post_` for consistency with other fields?

R/import_config.R
```diff
@@ -22,13 +22,40 @@ read_config <- function(path) {
...
+    transform = list(
+      list(
+        func = "func",
+        must_exist = TRUE
+      )
+    ),
+    load = list(
+      list(
+        func = "func",
+        must_exist = !info$load$automatic
```
richfitz, line 41: Is this quite right? If `func` is present then the pre and post bits might want to be definitely missing. Likewise with an automatic, load must be missing

R/import_config.R
```diff
@@ -88,31 +115,30 @@ set_missing_values <- function(info, field_name) {
   info
 }
 
-#' Read file fields of the dettl config yaml file.
+#' Read function fields of the dettl config yaml file.
 #'
 #' Reads the config file, and sources any file references.
 #'
-#' @param fields The name of the field.
+#' @param fields List of fields with functions and whther they have to exist.
```
richfitz: spelling of `whether`

R/import_config.R
```diff
@@ -88,31 +115,30 @@ set_missing_values <- function(info, field_name) {
...
-      assert_func_exists(config[[field]]$verification_queries, env)
-      config[[field]]$verification_queries <-
-        get0(config[[field]]$verification_queries,
-             envir = env,
-             mode = "function", inherits = FALSE
-        )
+  for (field in names(fields)) {
+    for (property in fields[[field]]) {
+      func_name <- config[[field]][[property$func]]
+      if (is.null(func_name) && property$must_exist) {
```
richfitz, line 134: Almost seems like it's a required/optional/allowed set of cases here that we need

R/load_runner.R
```diff
@@ -41,10 +42,17 @@ run_load <- function(con, load, extracted_data, transformed_data, test_queries,
   invisible(TRUE)
 }
 
-do_load <- function(con, load, extracted_data, transformed_data, path, test_file, test_queries,
-                    log_table, log_data, dry_run) {
+do_load <- function(con, load, extracted_data, transformed_data, path,
+                    test_file, test_queries, pre_load, post_load, log_table,
+                    log_data, dry_run) {
   before <- test_queries(con)
```
richfitz, line 50: While these run I think we might want to indicate what it is up to perhaps? So that if an error happens the user knows if it's in the pre, main or post load?

richfitz (changes requested):
Small things - I think that we might tidy up the required/allowed distinction in the configuration and make it more obvious how the load is going

R/load_runner.R
r-ash: Added some more messages saying what is being run.

R/import_config.R
r-ash: Fixed, thanks

R/dettl.R
r-ash: Updated that

R/import_config.R
r-ash: We already have a check that if automatic is TRUE then load should be missing. It is done in the call to `validate_load` which is a bit above this.

I've expanded the `validate_load` function to check that `pre_load` and `post_load` can only be defined if automatic load is being used.

richfitz (approved):

PR #39, opened 2020-02-14
Allow users to get extracted/transformed data if tests for that stage fail

r-ash:
Previously we were allowing the next stage to be run when the data from the previous stage existed

Now the data is saved on the object before the tests are run and we set a flag to true if the tests pass which the next stage uses to determine whether it can be run.

Diff is quite noisy couple of things adding lots of updated lines
* Lots of updated docs I think some due to newer version of roxygen on my laptop
* 2 new test "example-*" directories added to test that we can get extract/transform data when the tests fail. These are largely copies of existing test dir just with some small tweaks for the specific cases we want to test


R/dettl.R
```diff
@@ -31,6 +31,8 @@ DataImport <- R6::R6Class(
     load_ = NULL,
     extract_test_ = NULL,
     transform_test_ = NULL,
+    extract_passed = FALSE,
```
richfitz, line 36: What happens at present if the user does extract, transforms, then re-extracts, does `extract_passed` get re-FALSED (and the data get nulled?)

richfitz (changes requested):
Just a query about what happens with people iterating through things...

R/dettl.R
r-ash: We don't ever do that. We could have the `reload` function on the data import object reset these as currently that is how changes to sources can be used without rebuilding the object from scratch.

richfitz (changes requested):
tests failing due to docs mismatch

```
Codoc mismatches from documentation object 'run_transform':
run_transform
  Code: function(transform, extracted_data, extract_passed, path)
  Docs: function(transform, extracted_data, extract_passed)
  Argument names in code not in docs:
    path
```

r-ash:
I don't care about the warning on `DataImport` object for the being - this if fixed in later PR

richfitz (approved):

PR #41, opened 2020-02-17
Set up lintr running automatically on travis builds and fix warnings

r-ash:
Lots of changes here to fix lintr warnings.

Some warnings I've switched off because they are triggering on cases we want to ignore and every instance of this warning was a false positive e.g. variable names or function lengths


.lintr
```diff
@@ -0,0 +1,9 @@
+linters: with_defaults(
+    line_length_linter = NULL,
+    closed_curly_linter = NULL,
+    open_curly_linter = NULL,
+    object_name_linter = NULL,
+    object_length_linter = NULL
+  )
+exclusions: list("R/db.R" = 106)
```
richfitz: This feels fragile to me (any change to db.R before line 106) will break this.  So can we use one of the other exclusion approaches?  See `?lintr::exclude`.  It looks like this works:

```
#'
#' @keywords internal
# nolint start
verify_table <- function(con, table_name, table, identical_columns = FALSE,
                         context_info = "", solution_text = "") {
# nolint end
  if (!DBI::dbExistsTable(con, table_name)) {
    stop(sprintf(
```

richfitz (changes requested):
This looks good, thanks. I think this will be worth adding to other projects too.  One small suggestion for doing the exclusions more robustly

.lintr
r-ash: Fixed, thanks

richfitz (approved):

PR #42, opened 2020-02-21
Add docker support

richfitz:
Base docker image for dettl. We can build off this with specific montagu needs


weshinsley (approved):
Tested and talked through!

PR #43, opened 2020-02-21
vimc-3458: Add a script to simplify running things

richfitz:
This PR adds a small script to the docker image that we can use to automate running dettl from the command line (as we'll want to do from dettl).  The pattern is the same as used in orderly and other centre work with a small parse function that deals with docopt; this simplifies the testing a bit.


r-ash (approved):
Looks great

PR #44, opened 2020-02-21
Set working directory for transform & load and allow running import outside of a transaction

r-ash:
This PR will
* Set working directory of transform and load stages to be the import directory (now consistent over extract, transform and load)
* Allow users to turn off running the import inside a transaction
   * If the load throws an error it is still the responsibility of the `load` function to handle that and continue if it can - otherwise it will just exit
   * As part of this moved the call to sanitise the `dettl` part of the config into the `read_config` call


R/utils.R
```diff
@@ -2,6 +2,9 @@
   if (is.null(a)) b else a
 }
 
+`%?%` <- function(x, y) list(x = x, y = y)
```
richfitz: Please reformat to use braces:

```
`%?%` <- function(x, y) {
  list(x = x, y = y)
}
```

R/load_runner.R
```diff
@@ -23,20 +23,25 @@
...
+                     pre_load, post_load, path, test_file, transaction, dry_run,
+                     log_table, comment) {
   if (is.null(transformed_data)) {
     stop("Cannot run tests as no data has been transformed.")
   }
   log_data <- build_log_data(path, comment)
   verify_log_table(con, log_table, log_data)
   verify_first_run(con, log_table, log_data)
-  DBI::dbBegin(con)
+  if (transaction || dry_run) {
```
richfitz: perhaps do

```
use_transaction <- transaction || dry_run
```
above this and then use that value in both checks

R/load_runner.R
```diff
@@ -23,20 +23,25 @@
...
+  }
   withCallingHandlers(
     do_load(con, load, extracted_data, transformed_data, path, test_file,
-            test_queries, pre_load, post_load, log_table, log_data, dry_run),
+            test_queries, pre_load, post_load, log_table, log_data, transaction,
+            dry_run),
     error = function(e) {
-      DBI::dbRollback(con)
+      if (transaction || dry_run) {
+        DBI::dbRollback(con)
```
richfitz, line 44: Consider

```
if (!transaction) {
  message("ATTENTION: even though your load has failed, because you did not use a transaction, the database may have been modified")
}
```

or something similar

R/load_runner.R
```diff
@@ -45,22 +50,26 @@ run_load <- function(con, load, extracted_data, transformed_data, test_queries,
...
+    message("\t- Running load step")
+    load(transformed_data, con)
+    if (!is.null(post_load)) {
+      message("\t- Running post-load")
+      post_load(transformed_data, con)
+    }
+    message("\t- Running test queries after making changes")
+    after <- test_queries(con)
+  })
   test_path <- file.path(path, test_file)
```
richfitz: you could roll this section into the changed-dir above I think

R/transform_runner.R
```diff
@@ -20,7 +20,9 @@ run_transform <- function(con, transform, path, extracted_data,
     if (is.null(extracted_data)) {
       stop("Cannot run transform as no data has been extracted.")
     }
-    transformed_data <- transform(extracted_data)
+    withr::with_dir(path, {
+      transformed_data <- transform(extracted_data)
+    })
     verify_data(con, transformed_data, mode)
     if (!is.null(transform_test)) {
       message(sprintf("Running transform tests %s", transform_test))
```
richfitz: again, this could be done within the changed dir

richfitz (changes requested):
Small suggestions but this looks good. There's no ticket for this?

R/transform_runner.R
r-ash: Moved that into the withr block

R/load_runner.R
r-ash: Moved this into the withr block too - I agree we should be consistent about wd

R/load_runner.R
r-ash: Nice - added that

R/load_runner.R
r-ash: thanks

R/utils.R
r-ash: Have tidied that up

r-ash:
Sorry yeah didn't create a ticket thinking it would be some tiny update for working dir but then grew into its current state.

richfitz (approved):

PR #45, opened 2020-02-26
Fix bug where reloading dettl object was leaving stale connections mid transaction which blocked further db interaction

r-ash:
Note that we talked about adding a `finalize` method to the dettl object which would ensure DB connection was closed before garbage collection but this reliably causes a segfault when running the test suite. Including when I've tried to wrap the `dbDisconnect` call in a check if the connection is still valid.


R/dettl.R
```diff
@@ -42,7 +44,25 @@ DataImport <- R6::R6Class(
...
     db_name = NULL,
     mode = NULL,
-    transaction = TRUE
+    load_in_transaction = TRUE,
+
+    #' @description
+    #' Tidy up open connections, rolling back any active transactions
+    close_connection = function() {
+      if (private$transaction_active) {
+        warning("Rolling back active transaction")
```
richfitz: I am not sold on warning here - warnings get grouped together and are displayed at some future point. If you *really* want a warning, make it immediate, but TBH this looks more like a message (we're telling the user we're doing something) than anything else (I generally just avoid warnings)

R/dettl.R
```diff
@@ -42,7 +44,25 @@ DataImport <- R6::R6Class(
...
+        self$rollback_transaction()
+      }
+      ## Connection will always be null on first call to reload
+      if (!is.null(private$con)) {
+        tryCatch(
+          DBI::dbDisconnect(private$con),
+          error = function(e) {
+            ## Consume the error
```
richfitz: I think a notice here is still useful:

```
tryCatch(
  DBI::dbDisconnect(private$con),
  error = function(e) message("While disconnecting from db, ignored error:", e$message))
```

R/dettl.R
```diff
@@ -157,11 +178,37 @@ DataImport <- R6::R6Class(
       if (!force && !dry_run && !git_repo_is_clean(self$path)) {
         stop("Can't run load as repository has unstaged changes. Update git or run in dry-run mode.")
       }
-      run_load(private$con, private$load_, private$extracted_data,
-               private$transformed_data, private$test_queries,
-               private$load_pre_, private$load_post_, self$path,
-               private$load_test_, private$transaction, dry_run,
-               private$log_table, comment)
+
+      use_transaction <- private$load_in_transaction || dry_run
```
richfitz, line 210: This is the only place where `load_in_transaction` is used - perhaps drop it from the class and just use `dettl_config$dettl$transaction`

R/dettl.R
```diff
@@ -157,11 +178,37 @@ DataImport <- R6::R6Class(
...
+      if (use_transaction) {
+        self$begin_transaction()
+      }
+      message(
+        sprintf("Running load %s:",
+                use_transaction %?% "in a transaction" %:% "not in a transaction"))
```
richfitz: while this is cute, I wold have written this as:

```
      message(
        sprintf("Running load %s:",
                if (use_transaction) "in a transaction" else "not in a transaction"))
```

or even more concisely but perhaps less nicely

```
      message(
        sprintf("Running load %sin a transaction:",
                if (use_transaction) "" else "not"))
```

R/dettl.R
```diff
@@ -19,13 +19,15 @@
 #' import$transform()
 #' import$load()
 #'
-
+# nolint start
```
richfitz, line 22: This seems like a bit of a flaw in lintr that we can't just disable the complexity check here, and that it is flagging up the whole class and not the methods

richfitz (changes requested):

R/dettl.R
r-ash: Yeah bit of a shame it doesn't tell us where the actual problem is - we can disable the complexity check for the whole file

R/dettl.R
r-ash: Yeah that is a message really - have updated that

R/dettl.R
r-ash: Fixed

R/dettl.R
r-ash: We don't have the config at this point

R/dettl.R
r-ash: Tidied that up

richfitz (approved):

PR #49, opened 2020-02-28
Automatically detect ordering for automatic load

r-ash:
This PR will
* Order the transformed data before an automatic load is run (based upon the dependencies)

There are some fiddly things here mostly in `get_network_table` on ForeignKeyConstraint object. I've written notes in the code justifying what it is doing - I hope that is enough to make it make sense. Let me know otherwise and I will add more detail.


R/foreign_key_constraints.R
```diff
@@ -61,6 +64,47 @@ ForeignKeyConstraints <- R6::R6Class(
         }
       }
       is_serial && is_used_as_constraint
+    },
+
+    get_network_table = function(data) {
```
richfitz: I think we should lift this out into its own function that works on a character vector of table names.  That will be much easier to test, and look less daunting than a large method

tests/testthat/test-foreign-key-constraints.R
```diff
@@ -108,3 +112,41 @@ test_that("empty key constraints returns appropriate messages", {
 
   expect_false(keys$is_serial("table", "id"))
 })
+
+test_that("can get network table from foreign key constraints", {
```
richfitz: This seems heavier and less clear to me than something that is just pushing data around.  Why are mocks needed here?

richfitz (changes requested):
Two things that jump out, plus can we document or work through the situation of a loop in the graph?

tests/testthat/test-foreign-key-constraints.R
r-ash: To use the `ForeignKeyConstraints` object we'd have to setup a test db with the constraints which cover all the edge cases that the `constraint_table data` has in at the moment. That seemed more work than just mocking out the functions which use the db connection.

I could avoid this by having the `get_network_table` function on the object call a separate function which takes the table of constraints and the data and unit testing that instead.

R/foreign_key_constraints.R
r-ash: I can break this out into another function but we need the data for line 94 where we check for null or NA columns which we can ignore when building the upload order.

r-ash:
> Two things that jump out, plus can we document or work through the situation of a loop in the graph?

Fixed the 2 things notes. There is a test of cycles in the graph https://github.com/vimc/dettl/blob/fbcb3b408d85315532f3448f3fed9db7eaa5b7c8/tests/testthat/test-automatic-load.R#L179 and https://github.com/vimc/dettl/blob/fbcb3b408d85315532f3448f3fed9db7eaa5b7c8/tests/testthat/test-automatic-load.R#L227

I will add some more notes to the README though about this

r-ash:
I've added some notes to the vignette and added a test which explicitly checks that the automatic upload fails when we have a cycle and no columns are null

richfitz (approved):
Thanks - this looks great now

PR #52, opened 2020-09-04
VIMC-4254: Add start_time, end_time and duration to import log

r-ash:
This will require a db migration for science, production, annex (and uat?) so we should hold off on merging this until that is done. I will link to migration from here. Migration here https://github.com/vimc/montagu-db/pull/151

This PR will
* Remove `date` column from import log
* Add `start_time`, `end_time` and `duration` columns to import log
* Also tidy up some noisy test output and fix up some R CMD check stuff after coming back to this

I think teamcity is in a state so fine to ignore failing build


weshinsley (approved):
Looks good to me - nothing particular I can spot!

PR #53, opened 2020-09-09
Add some context to error message from dettl

r-ash:
See an example in the issue for the kind of error Katy has seen. This just adds some context to the error without losing the trace.


weshinsley (approved):
Looks fine to me. We should work out why teamcity is failing...

PR #54, opened 2020-09-09
Update vignette

r-ash:
This PR updates the vignette with info on
* Using modes
* Using `pre_load` and `post_load`
* Also reorganises vignette a bit and makes the automatic load more prominent

### DIFF
--- a/vignettes/dettl.Rmd
+++ b/vignettes/dettl.Rmd
@@ -31,9 +31,11 @@ ETL is a pattern for general copying of data from one or more sources to another
 
 **Load** - Step is responsible for loading the transformed data into the destination database. A set of post-load tests can be defined which will be run after the data is loaded into the database. If any of the tests fail the DB update will roll back, otherwise changes are committed.
 
-## File structure
+## Repository setup
 
-To define an import a directory structure as follows is required
+Dettl requires a config at the top level of a repository called `dettl_config.yml` which describes database connection information. Individual imports are then organised in separate subdirectories and must contain a config file `dettl.yml` which contains information relating to the import mode, files to be sourced and tests to be run. These subdirectories will be created automatically by using `dettl_new`.
+
+A typical repository might look like
 
 ```
 project_directory
@@ -45,9 +47,9 @@ project_directory
 |── dettl_config.yml
 ```
 
-Configuration files are required at 2 levels
+## dettl_config.yml
 
-* `dettl_config.yml` - Configuration at the project level used to configure DB connection information for any database which you wish to import to for each import subdirectory. Note that you must specify the `log_table` for each database. This is the table which import runs will be logged to.
+This needs to be setup manually and contains configuration at the project level used to configure DB connection information for any database which you wish to import to. Note that you must specify a `log_table` where run information is written after each import. See below for an example setup.
 
 ```yaml
 db:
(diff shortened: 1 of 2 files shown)


vignettes/dettl.Rmd
```diff
@@ -167,7 +132,9 @@ The extracted data can be inspected to ensure it looks as expected.
...
+### Transform
+
+Open the created `R/transform.R` file. Again you need to implement the templated `transform` function. Similarly to the extract stage this can delegate to other R functions if any files used are added to the `sources` block in the config. Similarly you should write some tests for the transformed data. Important to note is the form of the returned data should be as a named list of data frames. Where each list item represents a table from the database with name matching the name in the DB schema. And each column in the data frame should match an existing column in the database. After the transform step is run the user specified tests will be run automatically, in addition to these we test that the returned data conforms to the DB schema and fail if it does not do so. For example for our simple people example our transform code could be something like
```
weshinsley: In the middle - "Where each list" should be part of the previous sentence, not a new sentence...

vignettes/dettl.Rmd
```diff
@@ -199,7 +166,37 @@ The transformed data can be inspected to ensure it looks as expected.
...
+jobs <- data.frame(c(1, 2),
+                   c("researcher", "developer"),
+                   stringsAsFactors = FALSE)
+colnames(jobs) <- c("person", "job")
+knitr::kable(people, caption = "people")
+knitr::kable(jobs, caption = "jobs")
+```
+
+where `id` is the primary key of table `people` which is used as a foreign key for the `person` column of table `jobs`. To support linking between these tables in the transformed data use temporary unique ids to associate the new rows of data with each other. The tables themselves may already have entries with these primary keys used so when the automatic load is run the primary keys are updated when a new record is added. If you want to link to an existing entry in the database by its primary id then use that exact id as the field value.
+
```
weshinsley, line 198: Do you need to be specific here on what "temporary unique ids" to use? Negative numbers? Or just numbers that don't already exist as keys?

vignettes/dettl.Rmd
```diff
@@ -358,37 +355,66 @@ This will rollback any changes which have been made to the database before the c
...
+```
+mode: create
+```
+
+Create mode imports will skip some of the checks append imports do e.g. checks that the transformed data matches tables in the DB schema. The automatic load can still be used and will create each of the tables in the transformed data. Note that create and appending to tables cannot be mixed within an import.
+
+### I want to write a custom load
+
+Dettl allows writing a custom load if the automatic load does not achieve desired effect. You might want to use a custom load if you want to run an `UPDATE` query on the database. To do this update the `dettl.yml` for the report to use a user defined load
```
weshinsley: achieve *the* desired effect (perhaps?)

vignettes/dettl.Rmd
```diff
@@ -358,37 +355,66 @@ This will rollback any changes which have been made to the database before the c
...
-colnames(people) <- c("id", "name", "age", "height")
-jobs <- data.frame(c(1, 2),
-                   c("researcher", "developer"),
-                   stringsAsFactors = FALSE)
-colnames(jobs) <- c("person", "job")
-knitr::kable(people, caption = "people")
-knitr::kable(jobs, caption = "jobs")
+Note that the containing file must be in the list of sources for the import.
+Expect that this stage is only really a set of update queries to be run on the database implemented in R.
+
```
weshinsley, line 391: It might be good to say that if you use a custom load, then be careful not to specify numerical ids where the database would normally do that with a serial. Either leave the ids out and let the database do that itself, or if you require those ids in another place, then use an sql query with "RETURNING id" to fetch back the assigned serials. The key point is that doing a non-automatic custom load means dettl won't automatically fix your keys for you.... (and if you get this wrong like I did, you break the database!)

weshinsley (commented):
A few comments!

weshinsley (approved):
As this is a documentation one, I'll approve and let you go ahead and merge when happy!

PR #55, opened 2020-09-18
VIMC-3401: Update interface in preparation for adding sql only import

r-ash:
This PR will
* Rename `dettl_db_create_log_table` to `dettl_create_log_table`
* Remove exporting of `prepare_test_import`
* Add `dettl_run` function which can run multiple stages, can save output, run with path or import object
* Remove `dettl_run_extract`, `dettl_run_load`, `dettl_run_transform`
* Switch `force` arg on run to `allow_dirty_git` to be more clear about what it does

The main motivation for this is the switch from `dettl_run_extract`, `dettl_run_transform` and `dettl_run_load` to 1 function `dettl_run` this is because sql only import won't have stages and I just want to be able to call `dettl_load` to run it.

My plan after this is to
1. Add an interface/abstract class for an import which declares a `run` function and manages db connection
1. Extend this for the current type we have with specific impl for running and & yaml parsing for doing import (keeping the same functionality as the current code)
1. Add another extension to the interface which has specific impl for sql only imports


richfitz (approved):

richfitz:
Can you remove the TC build at some point? If you can't I can do it later.

PR #56, opened 2020-09-18
VIMC-4274: Add super class for import object

r-ash:
This PR will
* Add a super class which manages db connection
* A concrete implementation of it which doesn't change from previous for an R based import

This is in preparation for adding a sql only import and builds on #55


R/generic_import.R
```diff
@@ -0,0 +1,158 @@
+#' Generic import object.
+#'
+#' This shouldn't be initialised directly - this object manages transactions
+#' and the db connection and should be extended to implement actual imports.
+Import <- R6::R6Class(
```
r-ash, line 5: Decided to stick with the super class here over composition because I thought it does fit into the inheritance pattern but would be interested to get your opinion on that too? We can always separate it later if this starts becoming a bit of a god a class

R/generic_import.R
```diff
@@ -0,0 +1,158 @@
...
+        }
+      }
+      if (private$confirm) {
+        confirmed <- askYesNo(
+          sprintf(
+            "About to upload to database '%s' are you sure you want to proceed?",
+            private$db_name),
+          default = FALSE)
+        if (is.na(confirmed) || !confirmed) {
+          message("Not uploading to database.")
```
r-ash, line 102: Also I think could tidy this up, every other check throws an error but this one prints a message and returns. The return means the RImport class has a funny thing where it checks the result of the call to these checks and then returns itself if it is `FALSE` which feels a bit awkward. 

I thought about making this an error too but then that didn't feel good either because it is an expected user input. Any thoughts on nice way to make this a bit less awkward?

R/r_import.R
```diff
@@ -179,35 +152,13 @@ DataImport <- R6::R6Class(
...
       }
       message(sprintf("Running load %s", self$path))
-      if (!allow_dirty_git && !dry_run && !git_repo_is_clean(self$path)) {
-        stop("Can't run load as repository has unstaged changes. Update git or run in dry-run mode.")
-      }
 
-      use_transaction <- private$load_in_transaction || dry_run
+      use_transaction <- private$modify_in_transaction || dry_run
       if (use_transaction) {
         self$begin_transaction()
```
r-ash, line 169: I was also thinking about adding a function on the super class called like `with_transaction` or something which would take a block of code and manage the rollback or commit depending on outcome but thought being explicit here in the function might be more clear for reading.

r-ash:
I realise I create this under the wrong branch - see #57 for correct branch

PR #57, opened 2020-09-24
VIMC-4272: Refactor code in preparation for sql only import

r-ash:
This PR will

* Add a super class which manages db connection & load stage structure
* A concrete implementation of it which doesn't change from previous for an R based import
* A helper class for managing timing of import & writing log data
* Simplified `dettl_run` interface a bit - previously this would have hard to predict behaviour if calling with an import object and some other db_name param set. I have removed the ability to call it with an import object.

This is in preparation for adding a sql only import and builds on #55

After this plan is to write 2 PRs
* Add mode to import_log (atm is append or create I think) 
* Then add new sql only import mode


R/generic_import.R
```diff
@@ -0,0 +1,162 @@
+#' Generic import object.
+#'
+#' This shouldn't be initialised directly - this object manages transactions
+#' and the db connection and should be extended to implement actual imports.
+Import <- R6::R6Class(
```
r-ash, line 5: Decided to stick with the super class here over composition because I thought it does fit into the inheritance pattern but would be interested to get your opinion on that too? We can always separate it later if this starts becoming a bit of a god a class

R/generic_import.R
```diff
@@ -0,0 +1,162 @@
...
+        }
+      }
+      if (private$confirm) {
+        confirmed <- askYesNo(
+          sprintf(
+            "About to upload to database '%s' are you sure you want to proceed?",
+            private$db_name),
+          default = FALSE)
+        if (is.na(confirmed) || !confirmed) {
+          message("Not uploading to database.")
```
r-ash, line 192: Also I think could tidy this up, every other check throws an error but this one prints a message and returns. The return means the RImport class has a funny thing where it checks the result of the call to these checks and then returns itself if it is FALSE which feels a bit awkward.

I thought about making this an error too but then that didn't feel good either because it is an expected user input. Any thoughts on nice way to make this a bit less awkward?

R/r_import.R
```diff
@@ -171,43 +134,37 @@ DataImport <- R6::R6Class(
...
+      if (isFALSE(checks_passed)) {
+        return(invisible(FALSE))
       }
       message(sprintf("Running load %s", self$path))
-      if (!allow_dirty_git && !dry_run && !git_repo_is_clean(self$path)) {
-        stop("Can't run load as repository has unstaged changes. Update git or run in dry-run mode.")
-      }
 
-      use_transaction <- private$load_in_transaction || dry_run
+      use_transaction <- private$modify_in_transaction || dry_run
```
r-ash: I was also thinking about adding a function on the super class called like with_transaction or something which would take a block of code and manage the rollback or commit depending on outcome but thought being explicit here in the function might be more clear for reading.

r-ash:
Would be interested to see thoughts on this. Couple of things I played around with but decided against
* Adding another class which manages db connection and transactions only - I feel this ended up being too tightly coupled to what behaviour we want to do around when running dry_run, in/not in transaction and when tests pass/fail that the code would be less clear if the db transaction stuff was managed through a different class

Also `dettl_run` has specific args for an R import - the `stage` and `save. I still want to be able to pass those to `dettl_run` but not sure how this will work when we add other import types. Should I switch to using `...` args here and passing them through if the import takes them?

R/generic_import.R
richfitz: I am happy either way, really. It's fairly internally facing so we're free to move to composition later if we want

tests/testthat/helper-dettl.R
```diff
@@ -32,3 +32,8 @@ TestDataImport <- R6::R6Class(
     }
   )
 )
+
+mock_private <- function(obj, to_mock, value) {
+  unlockBinding(to_mock, obj$.__enclos_env__$private)
```
richfitz: What is this for? Generally `unlockBinding` is a function to avoid as the R people consider it "unsafe"

richfitz (changes requested):
Given the impact on the tests, this looks reasonable, but it's hard to see!

* The mock_private thing is a worry; in the short term we'll use some helpers but longer term separate file processing from object initialisation
* The steps (extract, transform, load, and pre/post load hooks) will move in another function to a generalised steps approach
* We'll implement the SQL version on top of this PR on a different branch to make sure that the interface is capable of coping

tests/testthat/helper-dettl.R
r-ash: Removed this now. We can add `transformed_data` for the test without call to `unlockBinding` and after the refactor the test which needed this for `pre` and `post` load can be removed as it essentially duplicated a test we had in `test_pre_post_load`

r-ash: Thinking about move to separating file processing from object initialisation I think maybe hard to separate out if we want to keep the `reload` call which reloads from file. The object will still have to take the file path and be able to read cfg from it.

richfitz (approved):

PR #58, opened 2020-10-30
VIMC-4278: Add sql only import

r-ash:
Builds upon #57 and is a proof that hopefully the refactor there should make this PR not too painful.

I think we should also add another PR to add import mode to the log maybe?


R/dettl_runner.R
```diff
@@ -10,12 +10,17 @@
 #'
 dettl <- function(path, db_name = NULL) {
   mode <- get_mode(path)
-  if (!(mode %in% c("create", "append"))) {
+  if ((mode %in% c("create", "append"))) {
+    import <- RImport$new(path, db_name)
+  } else if (mode == "sql") {
+    import <- SqlImport$new(path, db_name)
```
richfitz, line 16: can the SQL imports be create/append types? I guess I will find out below, though it does feel like these are different "axes" of types...

R/import_config.R
```diff
@@ -60,6 +60,54 @@ read_r_config <- function(path) {
...
+  sql_file <- file.path(path, info$load$sql)
+  assert_scalar_character(sql_file)
+  assert_file_exists(sql_file)
+  info$load$func <- get_sql_load(sql_file)
+  env <- load_sources(info$sources, path)
+  info <- read_function_fields(function_fields, info, env)
+  info$name <- basename(normalizePath(path, winslash = "/"))
+  info$path <- path
+  info$dettl$transaction <-
+    is.logical(info$dettl$transaction) %?% info$dettl$transaction %:% TRUE
```
richfitz: is there any reason you can't write this as:

```
info$dettl$transaction <-
  if (is.logical(info$dettl$transaction)) info$dettl$transaction else TRUE
```

tests/testthat/example_load_template/import.sql
```diff
@@ -0,0 +1,2 @@
+INSERT INTO people(name, age, height)
```
richfitz, line 1: this looks really nice

R/sql_import.R
```diff
@@ -0,0 +1,97 @@
...
+    #' Custom formatter for pretty printing object summary.
+    #' @param brief If TRUE then print a brief summary.
+    format = function(brief = FALSE) {
+      data_import_format(self, brief, class(self)[[1L]])
+    },
+
+    #' @description
+    #' Print help page for the object
+    help = function() {
+      utils::help(class(self)[[1L]], package = "dettl")
```
richfitz: this shows up as a coverage hole, but looks like the super class could provide it?

R/sql_import.R
```diff
@@ -0,0 +1,97 @@
...
+    #' Read and parse config from path.
+    read_config = function() {
+      private$import_config <- read_sql_config(self$path)
+    },
+
+    #' @description
+    #' Custom formatter for pretty printing object summary.
+    #' @param brief If TRUE then print a brief summary.
+    format = function(brief = FALSE) {
+      data_import_format(self, brief, class(self)[[1L]])
```
richfitz: could be provided by superclass?

richfitz (changes requested):
This looks great.

* I am not convinced that create/append/sql is the right taxonomy here; it feels like there are two axes that have been merged into one
* There are some minor coverage holes (10 lines total) but only 2 of which I can convince codecov to tell me about. These are worth investigating but don't sweat every last line if they're super tedious

R/sql_import.R
r-ash: Thanks yeah, that isn't needed. Have removed it from the `RImport` too.

R/import_config.R
r-ash: No reason - fixed that

R/dettl_runner.R
r-ash: Yeah perhaps language and mode.
Really create/append is a choice about how it can effect the db and sql vs others is purely how the import will be written

richfitz (approved):

PR #60, opened 2020-11-03
Fix postgres and vault connections in gh-actions

r-ash:
Couple of things possibly worth noting
* For postgres 11 you have to set password when starting docker container
* Setting the env var for vault which uses another env var is a bit of a faff, easiest way I could see was to do this in a separate step


richfitz (approved):
amazing, thank you

PR #61, opened 2020-11-04
VIMC-4279: Add language and mode to import log

r-ash:
Couple of things to note
* I've removed the tests which instantiate a "generic import" - I am viewing this as an abstract class which should never be instantiated and those tests were creating more problems than I think they were worth.
* Moved the language parsing stuff into the cfg reading function as this needed for all import types

I'll need to write migrations for the db schema


R/import_config.R
```diff
@@ -115,7 +116,13 @@ read_config_yml <- function(path) {
...
                      name = "Dettl configuration")
   info <- yaml_read(filename)
-  info$dettl$mode <- check_valid_mode(info$dettl$mode)
+  if (is.null(info$dettl$language)) {
+    ## Default use R as language
+    info$dettl$language <- "r"
+  }
+  assert_scalar_character(info$dettl$language, "import language")
+  ## we don't want to be case sensitive
+  info$dettl$language <- tolower(info$dettl$language)
```
richfitz, line 125: Do we not need to filter against a valid list here?

R/generic_import.R
```diff
@@ -136,20 +139,23 @@ Import <- R6::R6Class(
...
       private$confirm <- private$repo_config$db[[private$db_name]]$confirm
     },
 
+    # nocov start
+    ## Safe to ignore this in coverage as it doesn't and won't ever be run.
+    ## Viewing this as abstract impl of functions which a subclass must define
     #' @description
     #' Abstract impl should be overridden by subclass
     read_config = function() {
       NULL
```
richfitz: if you just drop the `NULL` here, then none of the nocov faff is needed

inst/sql/sqlite/create_log_table.sql
```diff
@@ -1,5 +1,7 @@
 CREATE TABLE dettl_import_log (
```
richfitz, line 1: I guess at some point we might look to including flyway migrations in the package, and some system for spitting those out so that it does not need to be manually copied over into the repos that use it. It's only us atm.

Having merged the montagu-db PR, I wonder if we might change this though, and have it write a "Schema version" into the table and refuse to change anything if the schema version does not match?

tests/testthat/test-load-runner.R
```diff
@@ -13,6 +13,8 @@ testthat::test_that("import fails if log table misconfigured", {
   invisible(DBI::dbExecute(con,
     "CREATE TABLE table_log (
       name   TEXT,
+      language TEXT,
```
richfitz, line 16: this is the sort of thing that a schema mismatch error would be nicer for:

"You are trying to import using dettl version 'x.y.x' but need to update your schema: see ?dettlschema for details"

richfitz (changes requested):
Some small comments and questions here

R/generic_import.R
r-ash: I tried that - it then causes an error in R6. Returns "attempt to apply non-function" e.g. when calling `import$load()`, even though the value of `load` has been overridden as a function. R6 treats methods and fields separately. It merges methods on class being instantaited & inherited ones, and the same for fields. Then adds methods to an env and then the fields to an env. Meaning if super class defines a field call `read_config` and subclass defines a method called `read_config` the field version will be available. See https://github.com/r-lib/R6/blob/5cd8acd0bc82488650460cbde30596052442d716/R/new.R#L116

inst/sql/sqlite/create_log_table.sql
r-ash: Added a ticket for this https://mrc-ide.myjetbrains.com/youtrack/issue/VIMC-4370

tests/testthat/test-load-runner.R
r-ash: Yeah that sounds much nicer, will add a note to the ticket above

R/generic_import.R
richfitz: I was thinking that `read_config = function() {}` would work here?

r-ash: :facepalm: yes sorry misunderstood first comment. That is nicer, made the switch and removed nocov stuff.

richfitz (approved):

PR #63, opened 2020-12-15
Enable sql only import to contain multiple statements

r-ash:
This does some basic cleaning of sql import statement file and then loops over all statements.

Bit of a pain RPostgres doesn't support this but see notes in code

### DIFF
--- a/R/sql_load.R
+++ b/R/sql_load.R
@@ -1,11 +1,24 @@
 get_sql_load <- function(path) {
-  statement <- paste(readLines(path), collapse = "\n")
+  ## It is not very nice that we have to read & parse this file to send it to
+  ## DB, ideally we could just execute the file directly on DB but DBI only
+  ## allows sending individual queries atm see
+  ## https://github.com/r-dbi/RPostgres/issues/213
+  sql <- readLines(path)
+  sql <- sql[!is_empty(sql)]
+  statement <- paste(sql, collapse = "\n")
+  statements <- unlist(strsplit(statement, ";", fixed = TRUE))
   function(con) {
     withCallingHandlers(
-      DBI::dbExecute(con, statement),
+      lapply(statements, function(cmd) {
+        DBI::dbExecute(con, cmd)
+      }),
       error = function(e) {
         e$message <- paste0("SQL error:\n", e$message)
         stop(e)
       })
   }
+}
+
+is_empty <- function(x) {
+  is.null(x) | is.na(x) | length(x) == 0 | trimws(x) == ""
 }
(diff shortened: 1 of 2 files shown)


R/sql_load.R
```diff
@@ -1,11 +1,24 @@
...
+      }),
       error = function(e) {
         e$message <- paste0("SQL error:\n", e$message)
         stop(e)
       })
   }
 }
+
+is_empty <- function(x) {
+  is.null(x) | is.na(x) | length(x) == 0 | trimws(x) == ""
```
richfitz: these should be "||"

R/sql_load.R
```diff
@@ -1,11 +1,24 @@
...
+  ## allows sending individual queries atm see
+  ## https://github.com/r-dbi/RPostgres/issues/213
+  sql <- readLines(path)
+  sql <- sql[!is_empty(sql)]
+  statement <- paste(sql, collapse = "\n")
+  statements <- unlist(strsplit(statement, ";", fixed = TRUE))
   function(con) {
     withCallingHandlers(
-      DBI::dbExecute(con, statement),
+      lapply(statements, function(cmd) {
```
richfitz: I think it'd be more idiomatic to use a for loop here as you're doing this for the side effect

R/sql_load.R
richfitz: but looking at how this is used on l7 I don't see why it's needed or why `sql[nzchar(sql)]` is insufficient

R/sql_load.R
```diff
@@ -1,11 +1,24 @@
 get_sql_load <- function(path) {
-  statement <- paste(readLines(path), collapse = "\n")
+  ## It is not very nice that we have to read & parse this file to send it to
+  ## DB, ideally we could just execute the file directly on DB but DBI only
+  ## allows sending individual queries atm see
+  ## https://github.com/r-dbi/RPostgres/issues/213
+  sql <- readLines(path)
+  sql <- sql[!is_empty(sql)]
+  statement <- paste(sql, collapse = "\n")
+  statements <- unlist(strsplit(statement, ";", fixed = TRUE))
```
richfitz: Having worked around this before I suggest you refactor this into one function to split your sql into chunks vs one to run it

richfitz (changes requested):

R/sql_load.R
r-ash: Yes - thanks

R/sql_load.R
r-ash: I want to vectorise it because `sql` is a vector on line 6 & 7. But `nzchar` is better - I just always forget that function exists.

richfitz: `is.null(x)` is scalar though so that would have been `||` otherwise it won't short circuit and you'll get odd answers from the following conditions:

```
x <- NULL
is.null(x) | is.na(x) | length(x) == 0 | trimws(x) == ""
# logical(0)
```

richfitz (approved):

PR #64, opened 2022-09-23
Add ability for dettl auto load to handle tables in different schemas

r-ash:
We want to put stochastic tables into a separate schema. This can't be done with dettl atm as the foreign key handling in the autoload cannot deal with tables in schema other than `public`. This updates the foreign key query to correctly label table with their namespaces.


R/db.R
```diff
@@ -140,7 +140,7 @@ verify_table <- function(con, table_name, table, additional_columns = NULL,
       if (!(col_name %in% db_col_names)) {
         stop(sprintf(
           "%s: Column '%s' is missing from db schema.",
-          context_info, col_name, table_name))
+          context_info, col_name))
```
r-ash, line 143: This was throwing a warning so just tidied up

tests/testthat/test-log-import.R
```diff
@@ -39,8 +39,8 @@ test_that("postgres log data is persisted", {
   expect_equal(postgres_data$name, "example")
   expect_equal(postgres_data$language, "r")
   expect_equal(postgres_data$mode, "append")
-  expect_equal(postgres_data$start_time, as.POSIXct(NA))
-  expect_equal(postgres_data$end_time, as.POSIXct(NA))
+  expect_equal(postgres_data$start_time, as.POSIXct(NA, tz = "UTC"))
+  expect_equal(postgres_data$end_time, as.POSIXct(NA, tz = "UTC"))
```
r-ash, line 43: This has started failing, assuming some change in DBI when reading time data or in expect_equal.

R/db.R
```diff
@@ -175,7 +175,8 @@ verify_table <- function(con, table_name, table, additional_columns = NULL,
 #' @keywords internal
 parse_sql_date <- function(con, date) {
   if (sql_dialect(con) == "sqlite") {
-    date <- as.POSIXct(date, origin = "1970-01-01", tz = "UTC")
+    date <- format(as.POSIXct(date, origin = "1970-01-01", tz = "UTC"),
+                   digits = 0)
```
r-ash, line 179: Also unrelated change, on devl ubuntu tests which check this were failing because the time was coming through as e.g. "16:34:23.12312354532424" where the test was not expecting any decimal points. This explicitly formats to remove any trailing digits as I think they are just precision issues from conversion anyway.

hillalex (approved):

